PR #2, opened 2022-02-03
New trunk angle implementation

adrien-berchet:
[imported from https://bbpgitlab.epfl.ch/neuromath/neurots/-/merge_requests/5]
This PR implements a refinement of trunk angle generation. It is able to detect and reproduce constraints between trunk directions from different neurite_types. For example, basal should not be 'too close' to apical, etc...

It can be used with [synthesis-workflow!9](https://bbpgitlab.epfl.ch/neuromath/synthesis-workflow/-/merge_requests/9) to produce validation plots.


arnaudon:
@lidakanari @adrien-berchet @eleftherioszisis , if you are happy with this implementation, which is not used by default, I'll add tests to get to better coverage. The plan would be to try to use that more in our next runs of synthesis before making it default here, so we keep the paper version until this new one is stable.

arnaudon:
@adrien-berchet , your commitlint thing is a pain, it checks all previous commits in the PR.... I squashed everything haha

arnaudon:
@adrien-berchet , it seems codecov needs to up updated, do you know how to do that?

adrien-berchet:
> @adrien-berchet , it seems codecov needs to up updated, do you know how to do that?

For an unknown reason the upload of the coverage report failed in the last job. I re-triggered the job and the upload succeeded this time.

arnaudon (approved):

arnaudon:
@adrien-berchet , this is you PR, so I can approve :P @lidakanari , let me know if/when you're happy with merging this! This new feature is off by default, so the behavior of the code will remain the same, unless specific options are used.

adrien-berchet:
> @adrien-berchet , this is you PR, so I can approve :P

:smile:

arnaudon:
Hello, shall we merge that? it's a bit of a pain to work on PR rebased on top of this one, it pollutes the diff quite a bit, and is prone to making mistakes. Thank you!

neurots/generate/soma.py
```diff
@@ -88,7 +88,7 @@ def point_from_trunk_direction(self, phi, theta):
     def orientation_from_point(self, point):
         """Return the orientation from the soma center to a point on the soma surface."""
         if np.allclose(point, self.center):
-            raise ValueError("Point overlaps with soma center.")
+            raise ValueError("Point overlaps with soma center.", point, self.center)
```
adrien-berchet, line 91:
```suggestion
            raise ValueError(f"Point overlaps with soma center at {self.center}.")
```

neurots/generate/orientations.py
```diff
@@ -277,29 +349,27 @@ def trunk_to_spherical_angles(trunk_angles, z_angles, phi_interval=None):
...
+    if form == "step":
+        scale, rate = params
+        return expit((angle - scale) / rate)
 
-    return phis, thetas
+    if form == "double_step":
+        scale_low, rate_low, scale_high, rate_high = params
+        return expit((angle - scale_low) / rate_low) + expit((-angle - scale_high) / rate_high)
+
+    raise NotImplementedError(f"form {form} not implemented")
```
adrien-berchet, line 372:
```suggestion
    raise NotImplementedError(f"The form '{form}' is not implemented")
```

neurots/generate/orientations.py
```diff
@@ -277,29 +349,27 @@ def trunk_to_spherical_angles(trunk_angles, z_angles, phi_interval=None):
...
+    """Prior uniform distribution on the sphere.
 
-    Args:
-        orientation (list[float]): The orientation vector.
-        trunk_absolute_angles (list[float]): The polar angles (phi in spherical coordinates).
-        z_angles (list[float]): The azimuthal angles (theta in spherical coordinates).
-
-    Returns:
-        tuple[numpy.ndarray[float], numpy.ndarray[float]]: The phi and theta angles.
+    If bias is tuple(direction, std), the sample will be along direction with normal std, improving
```
adrien-berchet, line 355: Missing args in docstring

neurots/extract_input/from_neurom.py
```diff
@@ -14,92 +14,178 @@
...
+    return _prob(angle) / quad(_prob, 0, np.pi)[0]
+
+
+def get_fit_prob_function(morph_class="PC", neurite_type=nm.BASAL_DENDRITE, params=None):
+    """Get probability function for trunk angles.
+
+    Args:
+        morph_class (str): morphological class (PC or IN)
+        neurite_type (nm.NeuriteType): type of neurite to consider
+        params (dict): is not None, can be used to overwrite the fit functions/bounds
```
adrien-berchet, line 78:
```suggestion
        params (dict): if not None, can be used to overwrite the fit functions/bounds
```

neurots/extract_input/from_neurom.py
```diff
@@ -14,92 +14,178 @@
...
+                "form": "double_step",
+                "bounds": ([-np.pi, 0.01, -np.pi, 0.01], [np.pi, 10, np.pi, 10]),
+            },
+            "axon": {"form": "step", "bounds": ([0.0, 0.01], [3.0, 10])},
+        },
+    }
+    if params is not None:
+        default_params.update(params)
 
+    bound = default_params[morph_class][neurite_type.name]["bounds"]
```
adrien-berchet, line 99: Maybe we should raise a clear exception if `morph_class` is not in `["PC", "IN"]`, because here it will just return a `KeyError` which might not be clear for the user.

neurots/extract_input/from_neurom.py
```diff
@@ -14,92 +14,178 @@
...
+    densities, bins = np.histogram(data, bins=bins, density=True)
+    return 0.5 * (bins[1:] + bins[:-1]), densities
 
-def soma_data(pop):
-    """Extract soma size.
+
+def trunk_vectors(morph, neurite_type, from_center_of_mass=True):
+    """This is neurom.get('trunk_vectors') but wrt to [0, 0, 0] if from_center_of_mass=False.
+
+    If one uses some.center (which is center of mass of soma points), it
```
adrien-berchet, line 114:
```suggestion
    If one uses `soma.center` (which is center of mass of soma points), it
```

neurots/extract_input/from_neurom.py
```diff
@@ -14,92 +14,178 @@
...
+def _get_hist(data, bins=50):
+    """Return density histogram with bin centers."""
+    densities, bins = np.histogram(data, bins=bins, density=True)
+    return 0.5 * (bins[1:] + bins[:-1]), densities
 
-def soma_data(pop):
-    """Extract soma size.
+
+def trunk_vectors(morph, neurite_type, from_center_of_mass=True):
+    """This is neurom.get('trunk_vectors') but wrt to [0, 0, 0] if from_center_of_mass=False.
```
adrien-berchet, line 112:
```suggestion
    """This is `neurom.get('trunk_vectors')` but wrt to `[0, 0, 0]` if `from_center_of_mass=False`.
```

neurots/extract_input/from_neurom.py
```diff
@@ -14,92 +14,178 @@
...
+
+def _trunk_neurite(pop, neurite_type, bins, params=None):
+    """Extract trunk angle data.
 
     Args:
         pop (neurom.core.population.Population): The given population.
+        neurite_type (neurom.core.types.NeuriteType): Consider only the neurites of this type.
+        bins (int or list[int] or str, optional): The bins to use (this pararmeter is passed to
+            :func:`numpy.histogram`).
+        params (dict): parameter to pass to fit functions to set the bounds
```
adrien-berchet, line 143:
```suggestion
        params (dict): parameters to pass to fit functions to set the bounds.
```

neurots/extract_input/from_neurom.py
```diff
@@ -168,6 +254,24 @@ def trunk_neurite(pop, neurite_type=nm.BASAL_DENDRITE, bins=30):
...
 
+def trunk_neurite(pop, neurite_type=nm.BASAL_DENDRITE, bins=30, params=None, method="simple"):
+    """Extract the trunk data for a specific tree type.
+
+    Args:
+        pop (neurom.core.population.Population): The given population.
+        neurite_type (neurom.core.types.NeuriteType): Consider only the neurites of this type.
+        bins (int or list[int] or str, optional): The bins to use (this pararmeter is passed to
+            :func:`numpy.histogram`).
+        method (str): Method to use
```
adrien-berchet, line 265:
```suggestion
        method (str): Method to use.
```

neurots/extract_input/input_parameters.py
```diff
@@ -15,7 +15,16 @@
...
-tmd_algos = ("tmd", "tmd_gradient", "tmd_apical")
+
+def _sort_neurite_types(neurite_types):
+    """Sort neurite types to comply with internal requirements.
+
+    For now, only apical should be first, so it is placed first, and other neurite_types can be
+    placed with respect to it.
+    """
+    if "apical" in neurite_types:
+        return list(["apical"] + [t for t in neurite_types if t != "apical"])
```
adrien-berchet, line 26:
```suggestion
        return ["apical"] + [t for t in neurite_types if t != "apical"]
```

neurots/extract_input/input_parameters.py
```diff
@@ -24,74 +33,59 @@ def parameters(
...
-    """Returns a default set of input parameters to be used as input for synthesis.
+    """Return a default set of input parameters to be used as input for synthesis.
 
     Args:
         origin (list[float]): The origin point.
         method (str): The method to use.
         neurite_types (list[str]): The neurite types to consider.
         feature (str): Use the specified TMD feature.
         diameter_parameters (dict or str): The parameters used for the diameters.
+        trunk_method (str): 'simple' for simple trunk method, or '3d_angles'
```
adrien-berchet, line 46:
```suggestion
        trunk_method (str): 'simple' for simple trunk method, or '3d_angles'.
```

adrien-berchet:
A few minor comments but it looks good to me.

arnaudon:
thanks @adrien-berchet !

neurots/extract_input/from_neurom.py
```diff
@@ -14,92 +14,184 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
+import warnings
```
lidakanari, line 17: I would say that anything related to the extract_input should be a separate PR (after the initial "trivial" one). Also in my opinion the distributions we extract should be the same independently of the algorithm. So instead of having options here, we should have a default that extracts all info "trunk_simple",  "trunk_with_probability" and what the algorithm will use should depend on the parameter file. This way we will only need to extract the distributions once, and then we can use them for all methods.

neurots/generate/orientations.py
```diff
@@ -16,8 +16,12 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import inspect
+import warnings
```
lidakanari, line 19: And then you can move these into a third PR to clean it up further.

lidakanari:
Here are some general comments. I propose to split into 3 PR:
1. trivial changes
2. related to input extraction
3. algorithms to grow the trunks accordingly

Also, it would be helpful to minimize the arbitrary fits etc as much as possible. If we can have a simple method to do the same process it will be easier to read and better to generate more accurate results.

arnaudon:
ok, I'm good to remove small changed, but 2 and 3 should go together IMO. I cannot test if my input makes sense if I cannot run the algo, no? Also changes are in different files, so it's same as 2 PR, no? Things are not on top of each others.

tests/test_neuron_functional.py
```diff
@@ -533,7 +533,7 @@ class TestBioRatL5Tpc1:
     def test_default(self):
         _test_full(
             "path_distances",
-            "bio_rat_L5_TPC_B.json",
+            "bio_rat_L5_TPC_B_distribution.json",
```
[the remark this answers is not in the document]
arnaudon: this was in another PR, rebase removed all these changes

neurots/extract_input/input_parameters.py
```diff
@@ -24,74 +35,69 @@ def parameters(
...
+
     input_parameters = {
-        "basal": {},
-        "apical": {},
-        "axon": {},
         "origin": list(origin),
-        "grow_types": neurite_types,
+        "grow_types": neurite_types
+        if trunk_method == "simple"
+        else _sort_neurite_types(neurite_types),
```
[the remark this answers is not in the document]
arnaudon: we cannot place a trunk that depends on another one being placed, so we need to force ordering.  This ordering is just later used to place trunks in this order, why would a user care about this?

neurots/extract_input/from_neurom.py
arnaudon: ok, I can then remove the option to have that, and use it by default. I wanted to hide it in this PR, until it is fully tested for real, etc... and don't loose computational time/disk space to save data we may not need. but as you prefer

neurots/generate/orientations.py
arnaudon: I would not split the input/algo changes, they go together, as I said below. it does not make sense to me to run extract without any algo to use the data, I would be 'blind' in the first PR

arnaudon:
> Also, it would be helpful to minimize the arbitrary fits etc as much as possible. If we can have a simple method to do the same process it will be easier to read and better to generate more accurate results.

I'm not sure I see how to d that. Fits are not arbitrary, they define the model and get rid of the noise/undersampling from the data. For having more accurate results, we need to see, depending on how we implement it.

arnaudon:
@adrien-berchet @eleftherioszisis , lint breaks in places not related to this PR, could one of you fix it in another one? thank you!

PR #24, opened 2022-03-02
Explicitely states where the examples and related data are located

adrien-berchet:


adrien-berchet:
Hi @alex4200 
In the OSS ticket you advised to mention where the examples and related data are actually located in the repo, so I add it in this PR.

PR #26, opened 2022-03-06
Update `pyproject.toml`

eleftherioszisis:
Add `build-system` entry as per https://setuptools.pypa.io/en/latest/build_meta.html

Reverted (See below for details):
Switch to preferred way to configure [setuptools_scm](https://setuptools.pypa.io/en/latest/build_meta.html)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,10 @@
+[build-system]
+requires = [
+    "setuptools>=42",
+    "wheel",
+]
+build-backend = "setuptools.build_meta"
+
 # BLACK
 [tool.black]
 line-length = 100


adrien-berchet (approved):

eleftherioszisis:
Unfortunately, I discovered that switching to the `pyproject.toml` configuration is not equivalent to using the `setup.py` as far as `setuptools_scm` is concerned.

setup.py commands do not install build dependencies from `pyproject.toml`, therefore if `python setup.py --version` is executed, it will silently return 0.0.0 using the `pyproject.toml` configuration unless the latest `setuptools_scm[toml]` and `setuptools` are already installed. Tox will also fail to get the version correctly for the same reason.

I will revert to the old way, as it is the most robust for the time being. Otherwise, we would need to add the dependencies in many places, such as in a `setup.cfg` and in `tox.ini`.

See for more details:
https://github.com/pypa/setuptools_scm/issues/386#issuecomment-575712244

adrien-berchet:
> Unfortunately, I discovered that switching to the `pyproject.toml` configuration is not equivalent to using the `setup.py` as far as `setuptools_scm` is concerned.

Ah :facepalm:

adrien-berchet (approved):

PR #25, opened 2022-03-06
Fix warnings

eleftherioszisis:

PR #27, opened 2022-03-07
Update scipy requirement because of QHullError

adrien-berchet:
Because of #25 we have to update requirements


eleftherioszisis:
I realized yesterday that this warning is only present in 1.8.0 .

Is it too restrictive to stick with the newest version or would it be better to live with the warning until there are a couple of scipy versions out before switching?

adrien-berchet:
I just ran the tests of `region-grower`, which runs a complete synthesis process, to ensure that `scipy >= 1.8` does not break anything and it worked. So I guess we can go with this version.

eleftherioszisis:
Unfortunately, 1.8 is not yet available in spack. I think we should revert to importing the error from `scipy.spatial.qhull` and live with the warning for a little longer.

https://github.com/BlueBrain/spack/blob/dadf42b62eb423e3f939f6668a4cc998e4e13487/var/spack/repos/builtin/packages/py-scipy/package.py#L17-L24

adrien-berchet:
Good point. We could add `scipy == 1.8` in spack though.
Or we can do a selective import like:
```python
try:
    from scipy.spatial import QhullError
except ImportError:
    from scipy.spatial.qhull import QhullError
```
so we have no warning and support all scipy versions.

eleftherioszisis:
> Good point. We could add `scipy == 1.8` in spack though. Or we can do a selective import like:
> 
> ```python
> try:
>     from scipy.spatial import QhullError
> except ImportError:
>     from scipy.spatial.qhull import QhullError
> ```
> 
> so we have no warning and support all scipy versions.

Would be nice to add a small comment as to why this is needed.

eleftherioszisis (approved):

lidakanari (approved):

PR #29, opened 2022-03-22
CI: Move black, codespell, isort, pycodestyle and pydocstyle from tox to pre-commit

adrien-berchet:
This commit also formats the files according to the new rules and add commitlint in CI.


arnaudon (approved):

adrien-berchet:
Hehe ;-P
Remember that you have to `pip install` it locally and call `pre-commit install` for optimal use. If you don't it will just behave as usual: the code is only checked in the CI.

adrien-berchet:
Ah and if you think the checks are too strict (I added a few new ones 😇), we can discuss it :-)

PR #36, opened 2022-05-13
proposal to fix bar length warning

arnaudon:


neurots/generate/algorithms/tmdgrower.py
```diff
@@ -56,7 +56,9 @@ def __init__(
         barSZ = np.min(get_lengths(self.ph_angles))
         stepSZ = self.params["step_size"]["norm"]["mean"]
         if stepSZ >= barSZ:
-            L.warning("Selected step size %f is too big for bars of size %f", stepSZ, barSZ)
+            if barSZ > 0.1:
+                self.params["step_size"]["norm"]["mean"] = barSZ
```
adrien-berchet, line 60: Since we check for `stepSZ >= barSZ`, we should set the value to `barSZ * factor` with `0 < factor < 1` to make the warning disappear (though you set it to `debug` so it will disappear anyway but the line 60 will be executed each time instead of just when it is needed).

eleftherioszisis (changes requested):
This is wrong on so many levels. It is not a fix, it internally overrides the input distributions, and uses a random heuristic to solve sth that is not an issue, but rather the expected behavior of the algorithm.

arnaudon:
Isn't it what you wrote here? 
```
By the way, reducing the segment length can fix that warning. Having big segment lengths results in big steps of path distances that "jump" some branching points. And if you want to capture all branching points, you could select your segment length distribution based on the smallest bar in the barcode iirc.
```
I just reduce the step size of the current section that has a small bar?  In fact, this is my question, I'm not so sure where this is changed, as there is a copy of this params in the constructor of the class, so I assumed it was valid only for one bar, which was to small. I assumed this would make a smaller step in a single place only,  but maybe not, that's why I just proposed it in case it may work!

eleftherioszisis:
> Isn't it what you wrote here?
> 
> ```
> By the way, reducing the segment length can fix that warning. Having big segment lengths results in big steps of path distances that "jump" some branching points. And if you want to capture all branching points, you could select your segment length distribution based on the smallest bar in the barcode iirc.
> ```
> 
> I just reduce the step size of the current section that has a small bar? In fact, this is my question, I'm not so sure where this is changed, as there is a copy of this params in the constructor of the class, so I assumed it was valid only for one bar, which was to small. I assumed this would make a smaller step in a single place only, but maybe not, that's why I just proposed it in case it may work!

Ah no, I meant that it is possible to change the segment length distribution from your input files, not change neurots to force it for everything.

arnaudon:
I'm checking the barcodes I have, which give me
```
Selected step size 1.000000 is too big for bars of size 0.757741
```
but it seems the smallest one using get_lenghts() is 3.5, so I'm not quite sure how to fix that in the barcodes distributions. But yeah, it would be better to fix that in extract module of neurots

arnaudon:
closing this, @lidakanari will make another one

PR #38, opened 2022-07-04
New extract_input for trunk angles

arnaudon:


neurots/extract_input/input_parameters.py
```diff
@@ -24,74 +35,69 @@ def parameters(
     neurite_types=("basal", "apical", "axon"),
     feature="path_distances",
     diameter_parameters=None,
+    trunk_method="simple",
 ):
-    """Returns a default set of input parameters to be used as input for synthesis.
+    """Return a default set of input parameters to be used as input for synthesis.
```
arnaudon, line 40: rebase issue

neurots/extract_input/input_parameters.py
```diff
@@ -24,74 +35,69 @@ def parameters(
...
 
     Args:
         origin (list[float]): The origin point.
         method (str): The method to use.
         neurite_types (list[str]): The neurite types to consider.
         feature (str): Use the specified TMD feature.
         diameter_parameters (dict or str): The parameters used for the diameters.
+        trunk_method (str): 'simple' for simple trunk method, or '3d_angles'.
```
arnaudon, line 48: to remove and append by default

neurots/extract_input/from_neurom.py
```diff
@@ -14,92 +14,184 @@
...
-                "max": opt_distr.params[1] + opt_distr.params[0],
-            }
-        }
-    elif opt_distr.type == "expon":
-        return {"expon": {"loc": opt_distr.params[0], "lambda": 1.0 / opt_distr.params[1]}}
-    return None
+    # this is the form of the expected parameter dict.
+    # The bounds correspond to the function's parameter min/max bounds for the fit
+    default_params = {
+        "PC": {
```
lidakanari, line 83: Why do we need specific sets of parameters in the "distribution" part? Also where do we get these from if the default is not used. maybe this can be simplified?

lidakanari, line 83: We have PC / IN specific properties. Can we simplify by "existence or absence of apical"?

neurots/extract_input/from_neurom.py
```diff
@@ -14,92 +14,184 @@
...
 
 import neurom as nm
 import numpy as np
 from neurom import stats
+from neurom.check.morphology_checks import has_apical_dendrite
+from neurom.core.types import tree_type_checker as is_type
+from scipy.integrate import quad
+from scipy.optimize import curve_fit
 
+from neurots.generate.orientations import prob_function
```
lidakanari, line 27: Needs to be added to this PR, it's now failing to extract.

lidakanari (changes requested):

neurots/extract_input/from_neurom.py
arnaudon: indeed I'll add now

neurots/extract_input/input_parameters.py
arnaudon: I cannot have both in the parameters, as this is where we decide which method we want to use

adrien-berchet:
> @adrien-berchet , stop with this useless comit lint, we'll squash everything to merge anyway

Yeah I am changing this to lint only the PR titles (because these titles are used as commit messages on the main branch), not the commit inside PRs.

arnaudon:
Ok, I removed from this one, we'll rebase once you merge this commitlint update

PR #42, opened 2022-08-19
Compatibility of NeuriteType with NeuroM/MorphIO

arnaudon:


neurots/generate/diametrizer.py
```diff
@@ -397,7 +407,9 @@ def build(
     and should only update the neuron object.
     """
     if neurite_types is None:
-        neurite_types = [STR_TO_TYPES.get(tree_type) for tree_type in ["apical", "basal"]]
+        neurite_types = [
+            getattr(SectionType, tree_type) for tree_type in ["apical_dendrite", "basal_dendrite"]
+        ]
```
adrien-berchet, line 412:
```suggestion
        neurite_types = [SectionType.apical_dendrite, SectionType.basal_dendrite]
```
Isn't it possible to just write this?

neurots/generate/diametrizer.py
```diff
@@ -18,11 +18,21 @@
...
+    if tree_type is None:
+        return list(neuron.iter())
+    return [i for i in neuron.iter() if i.type == tree_type]
+
+
+def root_section_filter(neuron, tree_type=None):
+    """Filters root sections according to type."""
+    if tree_type is None:
+        return list(neuron.root_sections)
+    return [i for i in neuron.root_sections if i.type == tree_type]
```
adrien-berchet, line 35: Where are these functions used?

arnaudon: later in this file, I copied them from morphio_utils.py which had only these 2 functions, only used in this module. So we simplify things a bit.

adrien-berchet: Ah it was moved from `neurots/morphio_utils.py`, I missed it.

arnaudon:
tests will pass once https://github.com/BlueBrain/TMD/pull/52 is merged and released

eleftherioszisis:
Nice. I am wondering, would it be possible to retain backward compatibility with the old types in the json configs?

arnaudon:
> Nice. I am wondering, would it be possible to retain backward compatibility with the old types in the json configs?

Probably, we could add an if in a few places to add _dendrite to basal and apical with a warning.

arnaudon:
I added a converter, so if we try to load old format files, it will convert them on the fly.

adrien-berchet:
But the JSON schemas should also accept both values, isn't it?

arnaudon:
ah yes, I didn't think of that, isnt't the schema checked after loading the json? I added the converter in the load_json function.

adrien-berchet:
Hmmm ok
But then we will have to use this `load_json` function in `region-grower` to handle this case because in `region-grower` we load the params and distrs and verify them only once before passing them to NeuroTS with `skip_validation=True`.

eleftherioszisis:
> I added a converter, so if we try to load old format files, it will convert them on the fly.

Perfect, thanks! Could you please also add a test ensuring compatibility with the old format?

If it's 100% certain that the changes allow to also work with the old format, then there is no need going up a major version.

arnaudon:
yes, good point, I'll make the converter a public function so we can reuse in region-grower, and make sure it's 100% working on old format with some test

have a nice week end

adrien-berchet:
O, perfect!

arnaudon:
I moved the converter to utils and added a test

neurots/extract_input/from_diameter.py
```diff
@@ -25,12 +25,18 @@
...
 
-from neurots.morphio_utils import NEUROM_TYPE_TO_STR
-from neurots.utils import _check
+from neurots.utils import NeuroTSError
 
 default_model = {"Rall_ratio": 3.0 / 2.0, "siblings_ratio": 1.0}
 
 
+def _check(data):
+    """Checks if data in dictionary are empty."""
```
adrien-berchet, line 34:
```suggestion
    """Check if data in dictionary are empty."""
```

neurots/generate/diametrizer.py
```diff
@@ -18,11 +18,21 @@
...
 import numpy as np
+from morphio import SectionType
 
-from neurots.morphio_utils import STR_TO_TYPES
-from neurots.morphio_utils import TYPE_TO_STR
-from neurots.morphio_utils import root_section_filter
-from neurots.morphio_utils import section_filter
+
+def section_filter(neuron, tree_type=None):
+    """Filters all sections according to type."""
```
adrien-berchet, line 25:
```suggestion
    """Filter all sections according to type."""
```

neurots/generate/diametrizer.py
```diff
@@ -18,11 +18,21 @@
...
+    if tree_type is None:
+        return list(neuron.iter())
+    return [i for i in neuron.iter() if i.type == tree_type]
+
+
+def root_section_filter(neuron, tree_type=None):
+    """Filters root sections according to type."""
```
adrien-berchet, line 32:
```suggestion
    """Filter root sections according to type."""
```

neurots/utils.py
```diff
@@ -41,8 +43,32 @@ def format_values(obj, decimals=None):
     return obj
 
 
-def _check(data):
-    """Checks if data in dictionary are empty."""
-    for key, val in data.items():
-        if len(val) == 0:
-            raise NeuroTSError(f"Empty distribution for diameter key: {key}")
+def convert_from_legacy_neurite_type(data):
+    """Convert legacy neurite type names, basal-> basal_dendrite and apical -> apical_dendrite."""
```
adrien-berchet, line 47:
```suggestion
    """Convert legacy neurite type names, basal -> basal_dendrite and apical -> apical_dendrite."""
```

neurots/utils.py
```diff
@@ -41,8 +43,32 @@ def format_values(obj, decimals=None):
     return obj
 
 
-def _check(data):
-    """Checks if data in dictionary are empty."""
-    for key, val in data.items():
-        if len(val) == 0:
-            raise NeuroTSError(f"Empty distribution for diameter key: {key}")
+def convert_from_legacy_neurite_type(data):
```
adrien-berchet, line 46: I think we should also raise a warning in this function when a legacy key is transformed, so users know they have to change their inputs before the next major version.

arnaudon: yes, done!

lidakanari:
I juts need to check once TMD is merged, but it looks good to me. So we should have old params / distr work as well? I think that's a great solution until everything is converted to the new format.

arnaudon:
ok, CI is happy, let me know if this is good for you, it will work on both versions and the converter can be used elsewhere.

neurots/utils.py
```diff
@@ -41,8 +44,37 @@ def format_values(obj, decimals=None):
...
+            data[key] = convert_from_legacy_neurite_type(data[key])
+
+        if isinstance(_data, list):
+            for i, d in enumerate(_data):
+                if d == "apical":
+                    _warn(d)
+                    data[key][i] = "apical_dendrite"
+                if d == "basal":
+                    _warn(d)
+                    data[key][i] = "basal_dendrite"
```
adrien-berchet, line 78: Why not using recursion here as for dict values?

arnaudon: You mean to deal with a possible list of lists? There are no such things in these files, no?

adrien-berchet: Yeah I don't think so indeed.

neurots/extract_input/input_parameters.py
```diff
@@ -79,19 +79,19 @@ def merged_params(data):
             }
         )
 
-    if "basal" in neurite_types:
-        input_parameters["basal"] = merged_params({"tree_type": 3})
+    if "basal_dendrite" in neurite_types:
```
lidakanari, line 82: if we want to preserve compatibility here we need "basal" or "basal_dendrite"

arnaudon: ah yes, I added a converter at the top and a warning

lidakanari:
ok, LGTM

lidakanari (approved):

PR #43, opened 2022-08-25
Kanari/improve length warning

lidakanari:
Improve definition of minimum bar length.
* New value "min_bar_length" is computed in distribution extraction.
* Distributions and schema are updated accordingly.
* Moved warning for "min_bar_length" inside validator check.

### DIFF
--- a/neurots/generate/algorithms/tmdgrower.py
+++ b/neurots/generate/algorithms/tmdgrower.py
@@ -16,9 +16,9 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 import copy
+import logging
 
 import numpy as np
-from tmd.Topology.analysis import get_lengths
 
 from neurots.generate.algorithms.abstractgrower import AbstractAlgo
 from neurots.generate.algorithms.barcode import Barcode
@@ -27,6 +27,8 @@ from neurots.generate.algorithms.common import bif_methods
 from neurots.generate.algorithms.common import section_data
 from neurots.morphmath import sample
 from neurots.morphmath.utils import norm
+
+L = logging.getLogger(__name__)
 
 
 class TMDAlgo(AbstractAlgo):
@@ -42,7 +44,14 @@ class TMDAlgo(AbstractAlgo):
     """
 
     def __init__(
-        self, input_data, params, start_point, context=None, random_generator=np.random, **_
+        self,
+        input_data,
+        params,
+        start_point,
+        skip_validation,
+        context=None,
+        random_generator=np.random,
+        **_,
     ):
         """TMD basic grower."""
         super().__init__(input_data, params, start_point, context)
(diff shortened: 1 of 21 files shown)


neurots/generate/algorithms/tmdgrower.py
```diff
@@ -45,23 +44,36 @@ class TMDAlgo(AbstractAlgo):
     """
 
     def __init__(
-        self, input_data, params, start_point, context=None, random_generator=np.random, **_
+        self,
+        input_data,
+        params,
+        start_point,
+        skip_validation=False,
```
arnaudon, line 51: we need to pass it via NeuronGrower, which is the class used in region-grower, otherwise we won't have access to this parameter

neurots/schemas/distributions.json
```diff
@@ -49,6 +49,10 @@
                     "$ref": "#/definitions/data_distr",
                     "description": "The data distribution used for the number of trees to synthesize"
                 },
+                "min_bar_length": {
+                    "type": "number",
```
adrien-berchet, line 53: It's also possible to add a min value because I guess it can not be `<0`
```suggestion
                    "type": "number",
                    "minimum": 0,
```
You can also add `"exclusiveMinimum": true` if it can not be `<=0`.

tests/test_json_schema.py
```diff
@@ -58,6 +58,7 @@ def test_json_schema():
                             },
                         },
                         "num_trees": {"type": "object"},
+                        "min_bar_length": {"type": "number"},
```
adrien-berchet, line 61: Why don't we use the actual schema here? I thought I changed this in an old previous PR :thinking:

adrien-berchet: Ah this change is stuck in #2 :smile:

lidakanari: not sure... you can change it in this one if you prefer

neurots/generate/algorithms/tmdgrower.py
lidakanari: This is a tmd-algo specific test (this value does not exist in other distributions except for the tmd ones). However, you can pass it down when you define the NeuronGrower, I am adding a test for that to see how it works (and increase coverage).

neurots/schemas/distributions.json
lidakanari: Good idea, thanks!

tests/test_json_schema.py
adrien-berchet: It depends when you think we will merge #2?

arnaudon: ah #2  will likely be abandonned, as we're trying to split it in 2 smaller PRs, maybe create another one with this schema updates?

adrien-berchet: Ah ok ok, I will create a new one then

arnaudon (approved):

neurots/schemas/distributions.json
```diff
@@ -49,6 +49,12 @@
...
+                    "minimum": 0,
+                    "minimum": 0,
```
adrien-berchet, line 55: I think one minimum is enough @arnaudon ;-p

arnaudon: ah lol

arnaudon, line 55:
```suggestion
```

lidakanari: :P

PR #46, opened 2022-08-30
Refactor: Apply Copier template

adrien-berchet:


adrien-berchet:
This PR is just about applying a repo template so that all our repos have the same structure and same generic Python stuffs (same config files, same format rules, etc.) and it will be easier to update them when a generic issue raises (like something to change in Github actions or something like that).

arnaudon (approved):

arnaudon:
Do you plan to add it to TMD as well?

adrien-berchet:
> Do you plan to add it to TMD as well?

I think so, except if you have any objection

PR #47, opened 2022-08-30
Use diameter-synthesis by default

arnaudon:
- Use diameter-synthesis latest algo as default instead of no diametrizer.
- To use no_diametrizer, specify 'no_diameters' instead of 'default'.
- We apply a post_growth operation of relabeling morphio section.id for easier use with external diametrizer, as any reloading to neurom for example will break the section.id mappings and will not allow one to easily change diamters of morphio object in place

### DIFF
--- a/examples/data/bio_distr.json
+++ b/examples/data/bio_distr.json
@@ -303,6 +303,6 @@
         }
     },
     "diameter": {
-        "method": "default"
+        "method": "no_diameters"
     }
 }
--- a/neurots/extract_input/input_distributions.py
+++ b/neurots/extract_input/input_distributions.py
@@ -29,6 +29,8 @@ from neurots.extract_input.from_TMD import persistent_homology_angles
 from neurots.utils import format_values
 from neurots.utils import neurite_type_warning
 
+from diameter_synthesis.build_models import build as build_diameter_models
+
 L = logging.getLogger(__name__)
 
 
@@ -85,20 +87,25 @@ def distributions(
 
     if diameter_input_morph is None:
         diameter_input_morph = filepath
-
-    if isinstance(diameter_model, str):
-        input_distributions["diameter"] = from_diameter.model(
-            load_morphologies(diameter_input_morph)
-        )
+    morphology = load_morphologies(diameter_input_morph)
+    if isinstance(diameter_model, str) and diameter_model.startswith('M'):
+        input_distributions["diameter"] = from_diameter.model(morphology)
         input_distributions["diameter"]["method"] = diameter_model
 
     elif hasattr(diameter_model, "__call__"):
-        input_distributions["diameter"] = diameter_model(load_morphologies(diameter_input_morph))
+        input_distributions["diameter"] = diameter_model(morphology)
         input_distributions["diameter"]["method"] = "external"
-    else:
+    elif (
+        isinstance(diameter_model, str) and diameter_model == "default"
+    ) or diameter_model is None:
+
         input_distributions["diameter"] = {}
         input_distributions["diameter"]["method"] = "default"
-        L.warning("No valid diameter model provided, so we will not generate a distribution.")
+        input_distributions["diameter"]["diameter"] = build_diameter_models(
+            morphology, config={"models": ["simpler"], "neurite_types": neurite_types}
+        )
+    else:
+        raise NotImplementedError(f"Diameter model {diameter_model} not understood")
 
     for neurite_type in neurite_types:
         if isinstance(feature, str):
--- a/neurots/extract_input/input_distributions.py
+++ b/neurots/extract_input/input_distributions.py
@@ -18,6 +18,7 @@
 import logging
 
 import tmd
+from diameter_synthesis.build_models import build as build_diameter_models
 from neurom import NeuriteType
 from neurom import load_morphologies
 
@@ -28,8 +29,6 @@ from neurots.extract_input.from_neurom import trunk_neurite
 from neurots.extract_input.from_TMD import persistent_homology_angles
 from neurots.utils import format_values
 from neurots.utils import neurite_type_warning
-
-from diameter_synthesis.build_models import build as build_diameter_models
 
 L = logging.getLogger(__name__)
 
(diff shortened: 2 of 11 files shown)


neurots/generate/grower.py
```diff
@@ -198,17 +200,20 @@ def grow(self):
...
     def _init_diametrizer(self, external_diametrizer=None):
         """Set a diametrizer function."""
-        if self.input_distributions["diameter"]["method"] == "default":
+        if self.input_distributions["diameter"]["method"] == "no_diameters":
             self._diametrize = lambda: None
             L.warning("No diametrizer provided, so neurons will have default diameters.")
         else:
             if self.input_distributions["diameter"]["method"] == "external":
-                if external_diametrizer is None:
-                    raise Exception("Please provide an external diametrizer!")
```
adrien-berchet, line 210: What happens now when `external_diametrizer is None`?

tests/test_neuron_functional.py
```diff
@@ -294,11 +290,11 @@ def test_breaker_of_tmd_algo():
     assert_array_equal(N.apical_sections, [33])
     assert_array_almost_equal(
         n.sections[169].points[-1],
-        np.array([-220.93813, -21.49141, -55.93323]),
+        np.array([4.5475516, 46.869442, 49.396336]),
```
adrien-berchet, line 293: Why changing the diameter model changes the result points? Is it because the diametrizer calls `np.random` and thus shifts the generated random numbers?

neurots/generate/grower.py
arnaudon: I noticed it was duplicate of a check in class __init__ just above

tests/test_neuron_functional.py
adrien-berchet: ah yeah ok.
Well, it would be nice to find the right ID so we can see that this commit changes section IDs but not the points (the current state implies that the points are change when using the new diametrizer while it is not the case).

neurots/generate/grower.py
adrien-berchet: Ah yes indeed!

tests/test_neuron_functional.py
arnaudon: I know, but I have no idea how to do that in a reasonable timeframe... if you have any idea, you're welcome!

adrien-berchet:
```python
    for num, sec in n.sections.items():
        if np.allclose(sec.points[-1], np.array([-220.93813, -21.49141, -55.93323]), atol=1e-3, rtol=1e-3):
            print(num)

    for num, sec in n.sections.items():
        if np.allclose(sec.points[-1], np.array([-17.31787, 151.4876, -6.67741]), atol=1e-3, rtol=1e-3):
            print(num)
```
did the trick ;-p

arnaudon: you are the best! thanks!

arnaudon:
@lidakanari ,are  you good to merge that, so we can release and bump version, or shall we release before?

arnaudon:
Shall we merge that?

adrien-berchet:
For me it looks good, I let @lidakanari approve.

examples/data/bio_distr.json
```diff
@@ -303,6 +303,6 @@
         }
     },
     "diameter": {
-        "method": "default"
+        "method": "no_diameters"
```
eleftherioszisis, line 306: Why not use json's `null` for the None case?

neurots/extract_input/input_distributions.py
```diff
@@ -85,20 +86,25 @@ def distributions(
 
     if diameter_input_morph is None:
         diameter_input_morph = filepath
-
-    if isinstance(diameter_model, str):
-        input_distributions["diameter"] = from_diameter.model(
-            load_morphologies(diameter_input_morph)
-        )
+    morphology = load_morphologies(diameter_input_morph)
+    if isinstance(diameter_model, str) and diameter_model.startswith("M"):
```
eleftherioszisis, line 90: Starts with "M"? Isn't it a bit arbitrary?

neurots/extract_input/input_distributions.py
```diff
@@ -85,20 +86,25 @@ def distributions(
...
-    else:
+    elif (
+        isinstance(diameter_model, str) and diameter_model == "default"
+    ) or diameter_model is None:
+
         input_distributions["diameter"] = {}
         input_distributions["diameter"]["method"] = "default"
-        L.warning("No valid diameter model provided, so we will not generate a distribution.")
+        input_distributions["diameter"]["diameter"] = build_diameter_models(
+            morphology, config={"models": ["simpler"], "neurite_types": neurite_types}
```
eleftherioszisis, line 104: Is there `simple` and then `simpler` and `simplest` ? :P
I would suggest instead of simple(r) to have something that is somewhat representative of what simple signifies here.

tests/test_neuron_functional.py
```diff
@@ -293,12 +289,12 @@ def test_breaker_of_tmd_algo():
 
     assert_array_equal(N.apical_sections, [33])
     assert_array_almost_equal(
-        n.sections[169].points[-1],
+        n.sections[118].points[-1],
```
eleftherioszisis, line 292: Why indices changed here?

neurots/extract_input/input_distributions.py
adrien-berchet: Ahah I had the same question :smile: 
I would also prefer a more meaningful name (I suggest `canconical-diametrizer` :stuck_out_tongue_winking_eye: ) but we would have to update `diameter-synthesis` first.

tests/test_neuron_functional.py
adrien-berchet: Because of this: https://github.com/BlueBrain/NeuroTS/pull/47#discussion_r959399932

neurots/extract_input/input_distributions.py
arnaudon: it's name of internal neurots diametrizers, M1, M2, etc...

neurots/extract_input/input_distributions.py
arnaudon: me too! if I find the person who named that function `simpler`....

examples/data/bio_distr.json
arnaudon: no idea :P

adrien-berchet (approved):

PR #49, opened 2022-09-12
3d trunk angles

arnaudon:


neurots/extract_input/input_distributions.py
```diff
@@ -77,6 +81,9 @@ def distributions(
             neurite_type_warning(neurite_type)
             neurite_types[i] = neurite_type + "_dendrite"
 
+    if trunk_method not in ("simple", "3d_angles"):
+        raise KeyError(f"trunk_method {trunk_method} not understood")
+
```
lidakanari, line 86: " ... not recognized, please choose between ()"

neurots/extract_input/input_parameters.py
```diff
@@ -47,6 +60,14 @@ def parameters(
             neurite_type_warning(neurite_type)
             neurite_types[i] = neurite_type + "_dendrite"
 
+    if method not in ("trunk", "tmd", "tmd_gradient", "tmd_apical"):
+        raise KeyError(f"Method {method} not understood.")
```
lidakanari, line 64: same as above, list allowed methods

neurots/extract_input/input_parameters.py
```diff
@@ -47,6 +60,14 @@ def parameters(
...
+    if trunk_method not in ("simple", "3d_angles"):
+        raise KeyError(f"trunk_method {trunk_method} not understood.")
```
lidakanari, line 66: same as above, list allowed methods

neurots/extract_input/input_parameters.py
```diff
@@ -103,6 +124,14 @@ def merged_params(data):
         if method == "tmd":
             input_parameters["apical_dendrite"]["growth_method"] = "tmd_apical"
 
+    if trunk_method == "3d_angles":
```
lidakanari, line 127: is it maybe possible to avoid having twice the same "if" maybe they can be put together?

neurots/generate/grower.py
```diff
@@ -298,58 +298,20 @@ def _convert_orientation2points(self, orientation, n_trees, distr, params):
...
-        for type_of_tree in tree_types:
+        return self.soma_grower.add_points_from_orientations(orientations)
 
+    def _simple_grow_trunks(self):
+        """Simple _grow_trunks function."""
+        for type_of_tree in self.input_parameters["grow_types"]:
             params = self.input_parameters[type_of_tree]
             distr = self.input_distributions[type_of_tree]
 
-            if legacy_mode:
```
lidakanari, line 332: @eleftherioszisis to check if it's ok to remove legacy

neurots/generate/orientations.py
```diff
@@ -56,6 +62,7 @@ def __init__(self, soma, parameters, distributions, context, rng):
 
         self._orientations = {}
         self._modes = self._collect_mode_methods()
+        self.pia_direction = np.array([0.0, 1.0, 0.0])
```
lidakanari, line 65: take this from the defined parameter?

neurots/schemas/distributions.json
```diff
@@ -74,7 +117,7 @@
                             "type": "null"
                         },
                         {
-                            "additionalProperties": false,
+                            "additionalProperties": true,
```
lidakanari, line 120: why this changed?

neurots/generate/grower.py
arnaudon: it's not, many test still rely on this legacy  mode, so I reverted

neurots/extract_input/input_parameters.py
arnaudon: it is!

neurots/generate/orientations.py
arnaudon: it is not in the parameter, unless we add it. In the other file, this direction is quite harcoded, I wonder if we should really hardcode it, or leave it free all the way?

neurots/schemas/distributions.json
arnaudon: I will clean these last, I was trying to quickly fix on the fly, but failed so far

neurots/generate/orientations.py
arnaudon: I've added it in utils.py as a variable we can import

arnaudon:
@lidakanari @adrien-berchet , I've clean up the main code, once we agree on it, I'll clean up the schema/tests so we can merge. I'm done with it for now.

neurots/generate/grower.py
```diff
@@ -367,6 +368,45 @@ def _grow_trunks(self):
...
+                        skip_validation=self.skip_validation,
+                        context=self.context,
+                        random_generator=self._rng,
+                    )
+                )
+
+    def _grow_trunks(self):
+        """Grow the trunks.
+
+        Two methods are available, depending on the data present in the input_parameters.
```
adrien-berchet, line 400:
```suggestion
        Two methods are available, depending on the data present in the `input_parameters`.
```

neurots/generate/grower.py
```diff
@@ -367,6 +368,45 @@ def _grow_trunks(self):
...
+        If no 3d_angles data is present, we grow trunks with :func:`_simple_grow_trunks` else
```
adrien-berchet, line 401:
```suggestion
        If no `3d_angles` entry is present, we grow trunks with :func:`_simple_grow_trunks` else
```

neurots/generate/orientations.py
```diff
@@ -197,6 +212,90 @@ def _mode_sample_pairwise_angles(self, _, tree_type):
...
+        return np.asarray(
+            [sample_spherical_unit_vectors(rng=self._rng) for _ in range(n_orientations)]
+        )
+
+    def _mode_normal_pia_constraint(self, values_dict, tree_type):
+        """Returns orientations using normal/exp distribution along a direction.
+
+        The `direction` value should be a 2-tuple, or a list of 2-tuples for multiple trunks.
+        The first value of the tuple is the angle wrt to pia ([0, 1, 0]) direction and the second
+        is the standard deviation of a normal distribution if angle>0 or scaling of exponential
```
adrien-berchet, line 227:
```suggestion
        is the standard deviation of a normal distribution if `angle>0` or scaling of exponential
```

neurots/generate/orientations.py
```diff
@@ -197,6 +212,90 @@ def _mode_sample_pairwise_angles(self, _, tree_type):
...
+        distribution if angle=0. As the resulting angle must be in [0, 2 * pi], we clip the
+        obtained angle and uniformaly sample the second angle to obtain a 3d direction.
```
adrien-berchet, line 229:
```suggestion
        obtained angle and uniformly sample the second angle to obtain a 3d direction.
```

neurots/generate/orientations.py
```diff
@@ -197,6 +212,90 @@ def _mode_sample_pairwise_angles(self, _, tree_type):
...
+                else:
+                    thetas.append(0)
+            else:
+                thetas.append(np.clip(np.random.normal(*direction), 0, np.pi))
+
+        phis = np.random.uniform(0, 2 * np.pi, len(values_dict["direction"]))
+        return spherical_angles_to_pia_orientations(phis, thetas)
+
+    def _mode_pia_constraint(self, _, tree_type):
+        """Create trunks from distribution of angles with pia ([0 , 1, 0]) direction.
```
adrien-berchet, line 250:
```suggestion
        """Create trunks from distribution of angles with pia (`[0 , 1, 0]`) direction.
```

neurots/generate/orientations.py
```diff
@@ -360,3 +459,190 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+
+def spherical_angles_to_pia_orientations(phis, thetas):
+    """Compute orientation from spherical angles where thetas are wirt to pia at [0, 1, 0].
+
+    Args:
+        phis (numpy.ndarray): Polar angles.
+        thetas (numpy.ndarray): Azimuthal angles.
+
+    Returns:
+        numpy.ndarray: The orientation vectors where each row correspnds to a phi-theta pair.
```
adrien-berchet, line 482:
```suggestion
        numpy.ndarray: The orientation vectors where each row corresponds to a phi-theta pair.
```

neurots/generate/orientations.py
```diff
@@ -360,3 +459,190 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+    """
+    assert PIA_DIRECTION == [0, 1, 0], "Global pia direction is not compatible"
```
adrien-berchet, line 484: This is only possible if the user deliberately updates the `neurots.generate.utils.PIA_DIRECTION`, right? In that case I'm not sure we have to check the value here. Maybe we could just ensure that it is not possible to update this value?

neurots/generate/orientations.py
```diff
@@ -360,3 +459,190 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+        (np.cos(phis) * np.sin(thetas), np.cos(thetas), np.sin(phis) * np.sin(thetas))
+    )
+
+
+def get_probability_function(form="step", with_density=True):
+    """Get probability functions to fit 3d trunk angles distributions.
+
+    Args:
+        form (str): form of the function, can be `flat`, `step` or `double_step`
+        with_density (bool): return the function with spherical density factor or not
```
adrien-berchet, line 495:
```suggestion
        form (str): Form of the function, can be `flat`, `step` or `double_step`.
        with_density (bool): Return the function with spherical density factor or not.
```

neurots/generate/orientations.py
```diff
@@ -360,3 +459,190 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+    Three forms of functions are available:
+    - `flat`: uniform flat distribution
+    - `step`: distribution with a single sigmoid :func:`scipy.special.expit`
+    - `double_step`: distribution with two opposite sigmoids :func:`scipy.special.expit`
+
+    Each sigmoid is  parametrized by a scale and a rate.
+
+    In practice, the `flat` function is used when no asymetry is present in the data, and the other
+    two are when an asymetry towards one direction, usually opposite to pia or apical,
+    or two directions, usually along and opposite to pia
```
adrien-berchet, line 506:
```suggestion
    Each sigmoid is parametrized by a scale and a rate.

    In practice, the `flat` function is used when no asymetry is present in the data, and the other
    two are when an asymmetry towards one direction, usually opposite to pia or apical,
    or two directions, usually along and opposite to pia.
```

neurots/generate/orientations.py
```diff
@@ -360,3 +459,190 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+                    expit((angle - scale_low) / rate_low) + expit((-angle - scale_high) / rate_high)
+                )
+
+            p = _prob(angle) / max(_prob(np.linspace(0, np.pi, 100)))
+            if with_density:
+                p *= np.sin(angle)
+            return np.clip(p, 0.0, 1.0)
+
+        return double_prob
+
```
adrien-berchet, line 548:
```suggestion
        return double_prob

    raise ValueError(f"The '{form}' value is unknown, it should be one of ['flat', 'step', 'double_step']")
```

neurots/generate/orientations.py
```diff
@@ -360,3 +459,190 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+        if form is not None:
+            if bounds is None:
+                bounds = fit_3d_angles_bounds[form]
+            if parameters["orientation"]["values"].get("params") is None:
+                return {"form": form, "bounds": bounds}
+    return None
+
+
+def fit_3d_angles(tmd_parameters, tmd_distributions):
+    """Fit functions to 3d_angles from tmd_distributions and save in tmd_parameters.
```
adrien-berchet, line 598:
```suggestion
    """Fit functions to 3d_angles from `tmd_distributions` and save in `tmd_parameters`.
```

neurots/generate/orientations.py
```diff
@@ -360,3 +459,190 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+    The update of tmd_parmeters is in plcae, and if fit parameters are already in tmd_parameters,
```
adrien-berchet, line 599:
```suggestion
    The update of `tmd_parmeters` is in place, and if fit parameters are already in `tmd_parameters`,
```

neurots/generate/orientations.py
```diff
@@ -360,3 +459,190 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+    the fit is skipped.
+
+    We return True if there is any 3d_angle data what was present in tmd_distributions, with or
+    without fit data, so this function can be used to detect if 3d_angles modes can be used
+    in synthesis.
```
adrien-berchet, line 604:
```suggestion
    We return `True` if there is any `3d_angle` data what was present in `tmd_distributions`, with or
    without fit data, so this function can be used to detect if `3d_angles` modes can be used
    in synthesis.
```

neurots/generate/orientations.py
```diff
@@ -360,3 +459,190 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+
+    Args:
+        tmd_parameters (dict): input parameters
+        tmd_distributions (dict): inputdistributions
```
adrien-berchet, line 608:
```suggestion
        tmd_parameters (dict): Input parameters.
        tmd_distributions (dict): Input distributions.
```

neurots/generate/orientations.py
```diff
@@ -360,3 +459,190 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+
+    Returns:
+        bool: True if 3d_angles mode can be used, False otherwise
```
adrien-berchet, line 611:
```suggestion
        bool: `True` if `3d_angles` mode can be used, `False` otherwise
```

neurots/extract_input/from_neurom.py
```diff
@@ -99,7 +102,77 @@ def soma_data(pop):
...
+        for n in nm.iter_neurites(morph, filt=is_type(neurite_type))
+    ]
+
+
+def trunk_neurite_3d_angles(pop, neurite_type, bins):
+    """Extract 3d trunk angle data.
+
+    We extract non-projected, or 3d angles between the pia/apical and any neurite trunk.
+    The direction of a trunk is defined from the vector between [0, 0, 0] and the root point,
+    to avoid any bias from non-spherical somata, where the soma center would be away from [0, 0, 0].
```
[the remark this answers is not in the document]
arnaudon: you mean in the docstring? Let see if this goes into neurom first

neurots/extract_input/input_parameters.py
```diff
@@ -47,6 +60,11 @@ def parameters(
             neurite_type_warning(neurite_type)
             neurite_types[i] = neurite_type + "_dendrite"
 
+    if trunk_method not in ("simple", "3d_angles"):
+        raise KeyError(
+            f"trunk_method {trunk_method} not understood, choose between {('simple', '3d_angles')}"
+        )
```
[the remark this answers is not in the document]
arnaudon: or remove it?

neurots/generate/grower.py
```diff
@@ -346,7 +347,7 @@ def _grow_trunks(self):
                 )
                 n_trees = len(orientations)
 
-                if type_of_tree == "basal_dendrite" and n_trees < 2:
+                if type_of_tree == "basal" and n_trees < 2:
```
arnaudon, line 350:
```suggestion
                if type_of_tree == "basal_dendrite" and n_trees < 2:
```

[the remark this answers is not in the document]
arnaudon: this is rebase issue I think

neurots/generate/orientations.py
```diff
@@ -197,6 +212,90 @@ def _mode_sample_pairwise_angles(self, _, tree_type):
...
+            propose = sample_spherical_unit_vectors(self._rng)
+            angle = nm.morphmath.angle_between_vectors(ref_dir, propose)
+            if self._rng.binomial(1, prob(angle, *params)):
+                return propose
+            n_try += 1
+        warnings.warn(
+            """We could not sample from distribution, so we take a random point.
+                      Consider checking the given probability distribution."""
+        )
+        return sample_spherical_unit_vectors(self._rng)
```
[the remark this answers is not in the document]
arnaudon: we don't know it, no?

neurots/generate/orientations.py
arnaudon: yes, I wasn't sure what's best here, @lida?

neurots/extract_input/from_neurom.py
```diff
@@ -99,7 +102,77 @@ def soma_data(pop):
...
+
+    If no apical dendrite is present, the entry `apical_3d_angles` will be absent.
+
+    Args:
+        pop (neurom.core.population.Population): The given population.
+        neurite_type (neurom.core.types.NeuriteType): Consider only the neurites of this type.
+        bins (int or list[int] or str, optional): The bins to use (this pararmeter is passed to
```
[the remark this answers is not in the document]
arnaudon: done

neurots/generate/grower.py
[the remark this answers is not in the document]
arnaudon: it's a bit to small to create a function, no?

neurots/extract_input/input_parameters.py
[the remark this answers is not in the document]
adrien-berchet: That's also possible indeed :)

neurots/generate/grower.py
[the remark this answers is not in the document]
adrien-berchet: Yeah and when the 'legacy' part will be removed it will not make sense to have a function for it. But having a function ensures that both cases are handled the same way. So as you prefer, but they at least have to be the same.

neurots/generate/orientations.py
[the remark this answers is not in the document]
adrien-berchet: Know what? That it would be better than a random angle you mean?

neurots/generate/grower.py
```diff
@@ -367,6 +368,45 @@ def _grow_trunks(self):
...
+
+    def _grow_trunks(self):
+        """Grow the trunks.
+
+        Two methods are available, depending on the data present in the `input_parameters`.
+        If no `3d_angles` entry is present, we grow trunks with :func:`_simple_grow_trunks` else
+        we fit the raw binned 3d angle data and apply :func:`_3d_angles_grow_trunks`.
+        """
+
+        if fit_3d_angles(self.input_parameters, self.input_distributions):
```
eleftherioszisis, line 405: You shouldn't mutate the attributes. They should be considered frozen. Make local copies if the initial data needs to be changed and pass it where needed. Debugging when the internal state of the instance changes based on how it is used is a nightmare.

I would dare go as far as converting the input_parameters and input_distribution into frozen dictionaries so that this is not allowed to occur :D

neurots/extract_input/from_neurom.py
```diff
@@ -99,13 +102,83 @@ def soma_data(pop):
...
+        """Return density histogram with bin centers."""
+        densities, _bins = np.histogram(data, bins=bins, density=True)
+        return densities, 0.5 * (_bins[1:] + _bins[:-1])
+
+    weights, _bins = _get_hist(pia_3d_angles)
+    data = {"pia_3d_angles": {"data": {"bins": _bins.tolist(), "weights": weights.tolist()}}}
+    if len(apical_3d_angles) > 0:
+        weights, _bins = _get_hist(apical_3d_angles)
+        data["apical_3d_angles"] = {"data": {"bins": _bins.tolist(), "weights": weights.tolist()}}
+    return {"trunk": data}
```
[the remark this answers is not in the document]
arnaudon: they can both be used, depending on how the user decides to use them

neurots/extract_input/from_neurom.py
```diff
@@ -168,6 +241,33 @@ def trunk_neurite(pop, neurite_type=nm.BASAL_DENDRITE, bins=30):
...
+
+    Returns:
+        dict: A dictionary with the trunk data.
+    """
+    trunk_data = trunk_neurite_simple(pop, neurite_type=neurite_type, bins=bins)
+    if method == "3d_angles":
+        trunk_data["trunk"].update(
+            trunk_neurite_3d_angles(pop, neurite_type=neurite_type, bins=bins)["trunk"]
+        )
+    return trunk_data
```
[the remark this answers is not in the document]
arnaudon: no no, we want to have both results, so user can use 3d_angles or previous algo, without having to re-extract this data

neurots/generate/orientations.py
```diff
@@ -197,6 +212,90 @@ def _mode_sample_pairwise_angles(self, _, tree_type):
...
+            with_density=False,
+        )
+        params = self._parameters[tree_type]["orientation"]["values"]["params"]
+        n_try = 0
+        while n_try < max_tries:
+            propose = sample_spherical_unit_vectors(self._rng)
+            angle = nm.morphmath.angle_between_vectors(ref_dir, propose)
+            if self._rng.binomial(1, prob(angle, *params)):
+                return propose
+            n_try += 1
```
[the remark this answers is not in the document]
arnaudon: I agree, I may do that in next PR with boundary, where accept/reject will be used in other places as well

neurots/generate/orientations.py
```diff
@@ -197,6 +212,90 @@ def _mode_sample_pairwise_angles(self, _, tree_type):
...
+        prob = get_probability_function(
+            form=self._parameters[tree_type]["orientation"]["values"]["form"],
+            with_density=False,
+        )
+        params = self._parameters[tree_type]["orientation"]["values"]["params"]
+        n_try = 0
+        while n_try < max_tries:
+            propose = sample_spherical_unit_vectors(self._rng)
+            angle = nm.morphmath.angle_between_vectors(ref_dir, propose)
+            if self._rng.binomial(1, prob(angle, *params)):
```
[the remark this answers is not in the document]
arnaudon: thanks!

neurots/extract_input/from_neurom.py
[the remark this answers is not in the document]
eleftherioszisis: Ah I see. It's clear now, thanks!

neurots/generate/grower.py
arnaudon: so you suggest fit_3d_angles to return a copy of input_parameters? (it's the only one mutated)

neurots/extract_input/input_distributions.py
```diff
@@ -77,6 +79,11 @@ def distributions(
             neurite_type_warning(neurite_type)
             neurite_types[i] = neurite_type + "_dendrite"
 
+    if trunk_method not in ("simple", "3d_angles"):
```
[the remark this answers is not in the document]
eleftherioszisis: For this use case, there is little to no difference.

But in general when you want to check sth against a list of objects, it's better using a set, instead of a tuple/list because you can check inclusion in O(1) instead of O(n).

neurots/extract_input/from_neurom.py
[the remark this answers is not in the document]
arnaudon: I added something, not sure if it's good docstring syntax

neurots/extract_input/from_neurom.py
```diff
@@ -99,13 +102,83 @@ def soma_data(pop):
...
+        vecs = trunk_vectors(morph, neurite_type=neurite_type)
+        pia_3d_angles += [nm.morphmath.angle_between_vectors(PIA_DIRECTION, vec) for vec in vecs]
+        if neurite_type.name != "apical_dendrite":
+            apical_ref_vec = trunk_vectors(morph, neurite_type=nm.APICAL_DENDRITE)
+            if len(apical_ref_vec) > 0:
+                apical_3d_angles += [
+                    nm.morphmath.angle_between_vectors(apical_ref_vec[0], vec) for vec in vecs
+                ]
+
+    def _get_hist(data):
```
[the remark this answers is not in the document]
arnaudon: it's used in other older part of the code, slightly differently, this may need to refactoring, but I don't want to touch it too much

neurots/generate/grower.py
arnaudon: I don't quite like it, I prefered my clean way with a True/False return, but if you prefer, here it is

neurots/generate/orientations.py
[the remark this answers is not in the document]
arnaudon: the angle we don't know, the l last one is the same as random

neurots/extract_input/from_neurom.py
```diff
@@ -99,13 +102,83 @@ def soma_data(pop):
     return {"size": transform_distr(ss)}
 
 
-def trunk_neurite(pop, neurite_type=nm.BASAL_DENDRITE, bins=30):
+def trunk_vectors(morph, neurite_type):
+    """This is `neurom.get('trunk_vectors')` but wrt to `[0, 0, 0]`."""
+    return [
+        nm.morphmath.vector(n.root_node.points[0], [0.0, 0.0, 0.0])
```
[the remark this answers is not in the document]
arnaudon: indeed, but if one provides non-centered morphologies to synthesis, it will give very weird angles haha. But it's hard to have a warning to detect that. We could see if somebody bumps into this issue badly at some point

neurots/generate/grower.py
```diff
@@ -402,7 +402,10 @@ def _grow_trunks(self):
         we fit the raw binned 3d angle data and apply :func:`_3d_angles_grow_trunks`.
         """
 
-        if fit_3d_angles(self.input_parameters, self.input_distributions):
+        with_3d, self.input_parameters = fit_3d_angles(
```
eleftherioszisis, line 405: Better but you are still overwriting the instance's `input_parameters`, which shouldn't change throughout its life.

You make a local copy of the parameters and pass it to the function that needs it.
```python
with_3d, input_parameters = fit_3d_angles(...)
if with_3d:
    self._3d_angles_grow_trunks(input_parameters)
```
However, it is peculiar that you need to change the parameters from within the grower. A more natural flow would be to try do the fitting before initializing the grower, passing to it the final parameters along with the option of which grow trunk algorithm to use.

neurots/extract_input/from_neurom.py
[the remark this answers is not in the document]
eleftherioszisis: I am not sure I understand why it would give weird angles if the soma center is always taken into account and the generation of points is done with respect to that.

neurots/generate/grower.py
arnaudon: the trick is that to get the fit one needs distributions, but parameters should not depend on distributions, if I understood @lidakanari  correctly. This means we either need an extra step before synthesis, or do it 'on the fly'. With this approach, we can do both: if we precompute the fit, this function will do nothing but check we can use 3d angle. In practice, we will always do that, where the fit will be done prior to running synthesis. This 'on the fly' approach is only for standalone use of neurots. 

I cannot pass the modified input_parameters to all the functions called later, it will be too much I think, no?

neurots/extract_input/from_neurom.py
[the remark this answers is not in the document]
arnaudon: it's for consistency, as we assign trunk angles first, then draw a soma, which center of mass will not be our original [0, 0, 0] position from which we've drawn angles, thus for example validation plots will have differences.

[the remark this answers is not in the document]
eleftherioszisis: Still, imo the reconstruction's center should be checked and raise an error if it's not centered at the origin. There is no guarantee that users will pass centered morphologies to the tool.

neurots/extract_input/from_neurom.py
```diff
@@ -99,13 +102,85 @@ def soma_data(pop):
...
+
+
+def trunk_neurite_3d_angles(pop, neurite_type, bins):
+    """Extract 3d trunk angle data.
+
+    We extract non-projected, or 3d angles between the pia/apical and any neurite trunk.
+    The direction of a trunk is defined from the vector between [0, 0, 0] and the root point,
+    to avoid any bias from non-spherical somata, where the soma center would be away from [0, 0, 0].
+
+    Warning: This is not compatible with the neurom `trunk_vectors` features.
```
eleftherioszisis, line 120: What's different? There is no way reusing existing code?

neurots/generate/grower.py
eleftherioszisis: This is not a matter of convenience, but rather one of correctness.

The `self.input_parameters` and `self.input_distributions` are, as their names indicate, inputs to the grower and are also used by classes that derive from the `NeuronGrower`, such as the `AstrocyteGrower` for instance. Therefore, we cannot modify them arbitrarily from inside the methods of the parent as they may result in unpredictable and hard-to-debug behavior.

As for where to add this step, I will leave this to @lidakanari to answer. I am not familiar with all the details of the algorithm to answer this correctly.

neurots/extract_input/from_neurom.py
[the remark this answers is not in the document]
arnaudon: this is no more

neurots/generate/grower.py
arnaudon: voila!

neurots/extract_input/input_parameters.py
```diff
@@ -103,6 +121,15 @@ def merged_params(data):
         if method == "tmd":
             input_parameters["apical_dendrite"]["growth_method"] = "tmd_apical"
 
+    if trunk_method == "3d_angles":
+        input_parameters["grow_types"] = _sort_neurite_types(neurite_types)
+        input_parameters["basal_dendrite"]["orientation"] = {"mode": "pia_constraint"}
+        if "apical_dendrite" in neurite_types:
+            input_parameters["apical_dendrite"]["orientation"] = {
+                "mode": "normal_pia_constraint",
+                "values": {"direction": [0.0, 0.0]},
```
lidakanari, line 130: direction [0,0] or [0,1,0]?

neurots/generate/orientations.py
```diff
@@ -360,3 +435,215 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+            return np.clip(p, 0.0, 1.0)
+
+        return double_prob
+
+    raise ValueError(
+        f"The '{form}' value is unknown, it should be one of ['flat', 'step', 'double_step']"
+    )
+
+
+def _fit_single_3d_angles(data, neurite_type, morph_class, fit_params=None):
```
lidakanari, line 519: morph_class is not very comprehensive, maybe apical_present instead or if this can be "True" "False" you can use has_apical

neurots/generate/orientations.py
```diff
@@ -360,3 +435,215 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+    to detect if `3d_angles` modes can be used in synthesis.
+
+    Args:
+        tmd_parameters (dict): Input parameters.
+        tmd_distributions (dict): Input distributions.
+
+    Returns:
+        tmd_parmeters with fit data if 3d_angles mode is found, else None
+    """
+    with_3d = False
```
lidakanari, line 582: in which case would this remain False? could this be a check implemented outside?

neurots/generate/orientations.py
```diff
@@ -360,3 +435,215 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+        bounds = parameters["orientation"]["values"].get("bounds")
+        if form is not None:
+            if bounds is None:
+                bounds = fit_3d_angles_bounds[form]
+            if parameters["orientation"]["values"].get("params") is None:
+                return {"form": form, "bounds": bounds}
+    return None
+
+
+def fit_3d_angles(tmd_parameters, tmd_distributions):
```
lidakanari, line 566: in general this function is not easy to follow. could it be simplified?

neurots/generate/orientations.py
```diff
@@ -360,3 +435,215 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+            if (
+                tmd_parameters[neurite_type]["orientation"].get("values") is None
+                or "params" not in tmd_parameters[neurite_type]["orientation"]["values"]
+            ):
+                data = tmd_distributions[neurite_type]["trunk"]["apical_3d_angles"]
+                new_tmd_parameters[neurite_type]["orientation"]["values"] = _fit_single_3d_angles(
+                    data["data"],
+                    neurite_type,
+                    morph_class,
+                    fit_params=_get_fit_params_from_input_parameters(tmd_parameters[neurite_type]),
```
lidakanari, line 606: this happens twice?

neurots/generate/orientations.py
```diff
@@ -360,3 +435,215 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+            if (
+                tmd_parameters[neurite_type]["orientation"].get("values") is None
+                or "params" not in tmd_parameters[neurite_type]["orientation"]["values"]
+            ):
+                data = tmd_distributions[neurite_type]["trunk"]["pia_3d_angles"]
+                new_tmd_parameters[neurite_type]["orientation"]["values"] = _fit_single_3d_angles(
+                    data["data"],
+                    neurite_type,
+                    morph_class,
+                    fit_params=_get_fit_params_from_input_parameters(tmd_parameters[neurite_type]),
```
lidakanari, line 620: also here?

neurots/generate/orientations.py
```diff
@@ -360,3 +435,215 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+    We also issue a warning so the user is aware that the provided distribution may have issues,
+    mostly related to large region of small probabilities.
+    """
+    prob = get_probability_function(
+        form=parameters[tree_type]["orientation"]["values"]["form"],
+        with_density=False,
+    )
+    params = parameters[tree_type]["orientation"]["values"]["params"]
+    n_try = 0
+    while n_try < max_tries:
```
lidakanari, line 639: so this could be unsuccessful? does this practically happens?

lidakanari (changes requested):

neurots/extract_input/input_parameters.py
arnaudon: it's [mean, sd] because it's 3d angle

neurots/generate/orientations.py
arnaudon: sure, but this name is from taxonomy file, which l looks like:
```
mtype   mClass  sClass
L1_DAC  INT INH
L1_HAC  INT INH
```

neurots/generate/orientations.py
arnaudon: if we are not in 3d_angle mode

neurots/generate/orientations.py
arnaudon: yes, I will try

neurots/generate/orientations.py
arnaudon: yes, if the fit is bad, below there is a warning explaining it

neurots/generate/orientations.py
arnaudon: yes, I simplified!

neurots/generate/orientations.py
arnaudon: simplified!

arnaudon:
@adrien-berchet @lidakanari , if this is what you had in mind, I'll fix the tests

neurots/preprocess/validity_checkers.py
```diff
@@ -39,6 +43,12 @@ def check_num_seg(params, distrs):
         )
 
 
+@register_global_validator()
+def validate(params, distrs):
```
adrien-berchet, line 47:
```suggestion
def check_input_validity(params, distrs):
```
Just to be consistent with our own convention.

neurots/preprocess/validity_checkers.py
```diff
@@ -1,11 +1,12 @@
...
-signature: `check_something(params, distrs)`. These functions should be very generic and should
-not depend on any context in which the related grower is used.
+The functions for validity checkers should have the signature: `check_something(params, distrs)`.
+These functions should be very generic and should not depend on any context in which the related
+grower is used.
 
 The checkers should be registered to be executed in the preprocess step using the
-`@register_validator` decorator.
+`@register_validator` decorator if they are applied per grow_type and growth_method, or with
+`@register_global_validator`  if they need the entire data.
```
adrien-berchet, line 9:
```suggestion
`@register_global_validator` if they need the entire data.
```

neurots/generate/grower.py
```diff
@@ -82,14 +82,14 @@ def __init__(
...
-        skip_validation=False,
+        skip_preprocessing=False,
         rng_or_seed=np.random,
         trunk_orientations_class=OrientationManager,
     ):
         """Constructor of the NeuronGrower class."""
         self.neuron = Morphology()
         self.context = context
-        self.skip_validation = skip_validation
+        self.skip_validation = skip_preprocessing
```
adrien-berchet, line 92:
```suggestion
        self.skip_preprocessing = skip_preprocessing
```
Let's be consistent everywhere

neurots/generate/grower.py
```diff
@@ -107,9 +107,10 @@ def __init__(
         self.input_distributions = _load_json(input_distributions)
 
         # Validate parameters and distributions
```
adrien-berchet, line 109:
```suggestion
        # Validate and preprocess parameters and distributions
```

neurots/generate/grower.py
```diff
@@ -107,9 +107,10 @@ def __init__(
...
-        if not skip_validation:
-            self.validate_params()
-            self.validate_distribs()
+        if not skip_preprocessing:
+            self.input_parameters, self.input_distributions = preprocess_inputs(
+                self.input_parameters, self.input_distributions
+            )
 
         # Consistency check between parameters and distributions
```
adrien-berchet, line 115: I realize that these checks on `metric` and `method` could also be moved to the validity checkers. But I will do this in another PR to not make this one too big.

adrien-berchet (commented):
The design looks good to me. For the rest I think it's ok but I will check it more deeply tomorrow.

neurots/generate/grower.py
arnaudon: added!

arnaudon:
@adrien-berchet , I simplified the validation of bar length as well, as it is done in the preprocessing, no need to redo it in the grower later

neurots/generate/orientations.py
```diff
@@ -16,15 +16,33 @@
...
+    "step": ([-np.pi, 0.1], [np.pi, 10]),
+}
+fit_3d_angles_params = {
+    "with_apical": {"basal_dendrite": {"form": "step", "bounds": fit_3d_angles_bounds["step"]}},
+    "without_apical": {"basal_dendrite": {"form": "flat", "bounds": []}},
+}
+_3d_angles_mapping = {
+    "apical_constraint": "apical_3d_angles",
+    "pia_constraint": "pia_3d_angles",
+}
```
adrien-berchet, line 45: I would use caps for these global variable to show that they should not be changed by the user.

neurots/generate/orientations.py
```diff
@@ -360,3 +438,216 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+    Args:
+        phis (numpy.ndarray): Polar angles.
+        thetas (numpy.ndarray): Azimuthal angles.
+
+    Returns:
+        numpy.ndarray: The orientation vectors where each row corresponds to a phi-theta pair.
+    """
+    return np.column_stack(
+        (np.cos(phis) * np.sin(thetas), np.cos(thetas), np.sin(phis) * np.sin(thetas))
+    )
```
adrien-berchet, line 455: I think this function should go in `morphmath.rotation`. But maybe with a warning in the docstring to clearly explain that this function is only relevant with angles wrt to pia direction.

neurots/generate/orientations.py
```diff
@@ -360,3 +438,216 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+                data["weights"],
+                bounds=_fit_params[morph_class][neurite_type]["bounds"],
+            )[0]
+        except RuntimeError:
+            warnings.warn("Cannot fit some trunk angles, we fallback to flat distribution")
+            form = "flat"
+            popt = np.array([])
+    else:
+        popt = np.array([])
+    return {"form": form, "params": popt.tolist()}
```
adrien-berchet, line 552:
```suggestion
            popt = curve_fit(
                function,
                data["bins"],
                data["weights"],
                bounds=_fit_params[morph_class][neurite_type]["bounds"],
            )[0].tolist()
        except RuntimeError:
            warnings.warn("Cannot fit some trunk angles, we fallback to flat distribution")
            form = "flat"
            popt = []
    else:
        popt = []
    return {"form": form, "params": popt}
```

neurots/generate/orientations.py
```diff
@@ -360,3 +438,216 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+    """Get parameter dict for fits from tmd_parameters."""
+    if "values" in parameters["orientation"] and parameters["orientation"]["values"] is not None:
+        form = parameters["orientation"]["values"].get("form")
+        bounds = parameters["orientation"]["values"].get("bounds")
+        if form is not None:
+            if bounds is None:
+                bounds = fit_3d_angles_bounds[form]
+            if parameters["orientation"]["values"].get("params") is None:
+                return {"form": form, "bounds": bounds}
+    return None
```
adrien-berchet, line 565:
```suggestion
    values = parameters["orientation"].get("values")
    if values is not None:
        form = values.get("form")
        if form is not None and values.get("params") is None:
            bounds = values.get("bounds", fit_3d_angles_bounds[form])
            return {"form": form, "bounds": deepcopy(bounds)}
    return None
```

neurots/generate/orientations.py
```diff
@@ -360,3 +438,216 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+
+        if (
+            tmd_parameters[neurite_type]["orientation"] is None
+            or "mode" not in tmd_parameters[neurite_type]["orientation"]
+        ):
+            if with_3d:
+                raise Exception("Only partial 3d_angle parameters are present")
+        elif tmd_parameters[neurite_type]["orientation"]["mode"] in _3d_angles_mapping:
+            with_3d = True
+    return with_3d
```
adrien-berchet, line 580:
```suggestion
    with_3d = False
    count_with_3d = 0
    for neurite_type in tmd_parameters["grow_types"]:
        if tmd_parameters[neurite_type]["orientation"].get("mode") in _3d_angles_mapping:
            with_3d = True
            count_with_3d += 1
    if with_3d and count_with_3d != len(tmd_parameters["grow_types"]):
        raise Exception("Only partial 3d_angle parameters are present")
    return with_3d
```
As far as I can see, with your implementation the exception is not raised if only the last `neurite_type` uses 3d angles.. I think it's not what you want, so this change will raise if at least one `neurite_type` is 3d but not all of them.

neurots/generate/orientations.py
```diff
@@ -360,3 +438,216 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+    We return `None` instead of copy of `tmd_parameters` if there is any `3d_angle` data that
+    was present in `tmd_distributions`, with or without fit data, so this function can be used
+    to detect if `3d_angles` modes can be used in synthesis.
+
+    Args:
+        tmd_parameters (dict): Input parameters.
+        tmd_distributions (dict): Input distributions.
+
+    Returns:
+        tmd_parmeters with fit data if 3d_angles mode is found, else None
```
adrien-berchet, line 597:
```suggestion
        `tmd_parmeters` with fit data if `3d_angles` mode is found, else None
```

neurots/generate/orientations.py
```diff
@@ -360,3 +438,216 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+                if _3d_angles_mapping[mode] not in tmd_distributions[neurite_type]["trunk"]:
+                    raise ValueError("No 3d angles found in distributions.")
+                tmd_parameters[neurite_type]["orientation"]["values"] = _fit_single_3d_angles(
+                    tmd_distributions[neurite_type]["trunk"][_3d_angles_mapping[mode]]["data"],
+                    neurite_type,
+                    morph_class,
+                    fit_params=_get_fit_params_from_input_parameters(tmd_parameters[neurite_type]),
+                )
+
+    return tmd_parameters
```
adrien-berchet, line 630:
```suggestion
    for neurite_type in tmd_parameters["grow_types"]:
        orientation = tmd_parameters[neurite_type]["orientation"]
        if orientation is None or "mode" not in orientation:
            continue

        mode = orientation["mode"]
        if mode in _3d_angles_mapping:
            if orientation.get("values", {}).get("params") is None:
                if _3d_angles_mapping[mode] not in tmd_distributions[neurite_type]["trunk"]:
                    raise ValueError("No 3d angles found in distributions.")
                orientation["values"] = _fit_single_3d_angles(
                    tmd_distributions[neurite_type]["trunk"][_3d_angles_mapping[mode]]["data"],
                    neurite_type,
                    morph_class,
                    fit_params=_get_fit_params_from_input_parameters(tmd_parameters[neurite_type]),
                )

    return tmd_parameters
```

neurots/generate/orientations.py
```diff
@@ -360,3 +438,216 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+
+
+def _sample_trunk_from_3d_angle(parameters, rng, tree_type, ref_dir, max_tries=100):
+    """Sample trunk directions from fit of distribution of 3d_angles wrt to ref_dir.
```
adrien-berchet, line 634:
```suggestion
    """Sample trunk directions from fit of distribution of `3d_angles` wrt to `ref_dir`.
```

neurots/generate/orientations.py
```diff
@@ -360,3 +438,216 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+        bounds = parameters["orientation"]["values"].get("bounds")
+        if form is not None:
+            if bounds is None:
+                bounds = fit_3d_angles_bounds[form]
+            if parameters["orientation"]["values"].get("params") is None:
+                return {"form": form, "bounds": bounds}
+    return None
+
+
+def check_3d_angles(tmd_parameters):
```
adrien-berchet, line 568: missing docstring

neurots/generate/orientations.py
```diff
@@ -360,3 +438,216 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+            warnings.warn("Cannot fit some trunk angles, we fallback to flat distribution")
+            form = "flat"
+            popt = np.array([])
+    else:
+        popt = np.array([])
+    return {"form": form, "params": popt.tolist()}
+
+
+def _get_fit_params_from_input_parameters(parameters):
+    """Get parameter dict for fits from tmd_parameters."""
```
adrien-berchet, line 556:
```suggestion
    """Get parameter dict for fits from `tmd_parameters`."""
```

neurots/generate/orientations.py
```diff
@@ -360,3 +438,216 @@ def compute_interval_n_tree(soma, n_trees, rng=np.random):
...
+
+        return double_prob
+
+    raise ValueError(
+        f"The '{form}' value is unknown, it should be one of ['flat', 'step', 'double_step']"
+    )
+
+
+def _fit_single_3d_angles(data, neurite_type, morph_class, fit_params=None):
+    """Fit function to distribution of 3d angles for a neurite_type.
```
adrien-berchet, line 523:
```suggestion
    """Fit function to distribution of 3d angles for a `neurite_type`.
```

neurots/morphmath/sample.py
```diff
@@ -145,3 +145,13 @@ def ph(phs, random_generator=np.random):
     """Samples randomly a persistence diagram from the input distribution."""
     index = random_generator.choice(len(phs))
     return phs[index]
+
+
+def sample_spherical_unit_vectors(rng):
```
adrien-berchet, line 150: Isn't it the same as `morphmath.utils.get_random.point`? I mean, it's not the same method but the objective is the same, right?

neurots/schemas/parameters.json
```diff
@@ -107,7 +107,7 @@
                                             "additionalProperties": false,
                                             "required": ["orientations"]
                                         },
-                                        {
+                                         {
```
adrien-berchet, line 110:
```suggestion
                                        {
```

neurots/schemas/distributions.json
adrien-berchet: Let's keep this unresolved for now to not forget.

neurots/astrocyte/grower.py
```diff
@@ -100,15 +100,13 @@ def __init__(
         input_distributions,
         context,
         external_diametrizer=None,
-        skip_validation=False,
```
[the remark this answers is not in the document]
arnaudon: indeed, my bad

neurots/generate/orientations.py
arnaudon: well, there is spherical_angles_orientation in this module, that's why it is here

neurots/generate/orientations.py
arnaudon: yes, thanks!

neurots/morphmath/sample.py
arnaudon: yes, it's probably the same, but mine is cleaner/faster ;P

neurots/schemas/distributions.json
arnaudon: ah yes, I added the missing entries in the schema

neurots/extract_input/input_parameters.py
```diff
@@ -20,13 +20,24 @@
 tmd_algos = ("tmd", "tmd_gradient", "tmd_apical")
 
 
+def _sort_neurite_types(neurite_types):
+    """Sort neurite types to comply with internal requirements.
+
+    For now, only apical should be first, so it is placed first, and other `neurite_types` can be
+    placed with respect to it.
+    """
```
lidakanari, line 28: any restrictions for the order of axons / basals?

neurots/morphmath/sample.py
```diff
@@ -145,3 +145,13 @@ def ph(phs, random_generator=np.random):
...
+    """Sample a point uniformly on the sphere.
+
+    Args:
+        rng: random number generator
+    """
+    x = rng.normal(0, 1, 3)
```
lidakanari, line 156: I *think* this is wrong. If you want to uniformly sample on a sphere you need to sample the angles, not x-y-z. Unless what you need to do here is not the same as randomly sample one point on a sphere :) please clarify, as this is an important point

lidakanari: Not for the code, but for the logic of the issue is you sample x,y,z instead: https://www.bogotobogo.com/Algorithms/uniform_distribution_sphere.php

neurots/extract_input/input_parameters.py
arnaudon: ah I forgot to remove that, it was bothering people, so I sort it externally now

neurots/morphmath/sample.py
arnaudon: what is wrong exactly? I want x, y, z on a sphere, uniformly sampled, which is achieved by this 'trick'. It's faster than sampling angles and computing trigonometrics functions.

arnaudon: eq (16) here: https://mathworld.wolfram.com/SpherePointPicking.html

lidakanari: let's talk in person, this is a hard concept but it's not correct to compute this way (I'll send plot in slack)

lidakanari: got it, works this way as well. maybe I should replace the other implementation :+1:

arnaudon:
I think I bumped to the wrong version

arnaudon:
ah no, so i don't know that this is..

adrien-berchet:
It's because of numpy, it's because you remove the `skip_validation` parameter from the `TreeGrowerSpaceColonization` constructor but in the test the `skip_validation` parameter is still given, so the parameters are shifted and thus the context is passed as a random number generator.

arnaudon:
Finally! I can't spot what is not covered, maybe codecov needs some time to be up to date.

arnaudon:
ALL GREEEEN!!!!!!

adrien-berchet:
> ALL GREEEEN!!!!!!

Yay :partying_face:

neurots/schemas/parameters.json
```diff
@@ -117,6 +117,38 @@
...
                                             "required": ["primary_orientation"]
+                                         },
+                                          {
+                                            "type": "object",
+                                            "properties": {
+                                                "direction": {
+                                                    "items": {
+                                                        "maxItems": 2,
+                                                        "minItems": 2,
+                                                        "description": "The orientation 3d angle (mean, sd)"
```
adrien-berchet, line 128: These entries should one level up (outside of the `items` object).

neurots/schemas/distributions.json
```diff
@@ -38,6 +43,44 @@
             "title": "Empty distribution",
             "type": "object"
         },
+        "neurite_diameter": {
+            "additionalProperties": false,
+            "description": "The diameter distributions of a type of neurite",
+            "properties": {
+                "Rall_ratio": {
+                    "type": "number"
```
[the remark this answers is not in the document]
arnaudon: I'm not sure why this is in this PR, it's not about diameters. @adrien-berchet I tried to rebase on main, but the merge is weird, I see you merged main into this branch recently?

neurots/schemas/parameters.json
```diff
@@ -117,6 +117,38 @@
                                             },
                                             "additionalProperties": false,
                                             "required": ["primary_orientation"]
+                                         },
+                                          {
+                                            "type": "object",
+                                            "properties": {
+                                                "direction": {
```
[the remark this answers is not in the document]
arnaudon: in the code I have direction=[0, 1] for example, would this work?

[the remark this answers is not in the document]
adrien-berchet: You would have to tranform the dict into list but I think it would be more explicit for the user to enter an object with named values instead of a list of 2 values with implicit meaning (mean and std respectively).

neurots/schemas/distributions.json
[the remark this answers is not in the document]
adrien-berchet: Ah maybe it's rebase issue. This block was moved but I don't know why. Mayb you have an auto-formatter for JSON in you editor? Or maybe it's a precommit hook?
Anyway, we can improve this schema in another PR.

neurots/schemas/parameters.json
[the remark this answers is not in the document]
arnaudon: yes, you're right, I can try that

neurots/schemas/distributions.json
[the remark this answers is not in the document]
arnaudon: ah alright, no i don't format these

neurots/schemas/parameters.json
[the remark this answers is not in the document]
arnaudon: done!

neurots/schemas/parameters.json
```diff
@@ -117,22 +117,42 @@
...
+                                                    "properties": {
+                                                        "mean": {
+                                                            "oneOf": [
+                                                                {
+                                                                    "type": "number",
+                                                                    "minimum": 0,
+                                                                    "maximum": 3.1416
+                                                                },
+                                                                {
+                                                                    "type": "array"
```
adrien-berchet, line 135: When it is an array, what is it's structure? Is it an array of numbers between 0 and pi? If yes we should also constrain the items of this array.

arnaudon: yes, list of such numbers, when there are more than one apical. How to do that?

arnaudon: @adrien-berchet, like that?

adrien-berchet (approved):
LGTM!

neurots/schemas/parameters.json
adrien-berchet: Yep, thanks :)

lidakanari:
Finally double checking and all is good from my side to merge this one. @eleftherioszisis if you want to check for astrocytes please provide feedback for this, otherwise we are ok to merge this one.

lidakanari (approved):

arnaudon:
great thanks! I'll then make a minor release of neurots, so we can merge the dependencies

PR #52, opened 2022-10-04
Feat: Rework input validation process

adrien-berchet:


neurots/preprocess/__init__.py
```diff
@@ -0,0 +1,56 @@
+"""Functions used for preprocessing."""
+from collections import defaultdict
+
+from neurots.generate.algorithms import tmdgrower
+from neurots.validator import validate_neuron_distribs
+from neurots.validator import validate_neuron_params
+
+_preprocess_functions = defaultdict(set)
```
eleftherioszisis, line 8: Minor: I would capitalize this as a global entity.

eleftherioszisis: And please don't add code to `__init__.py` modules if it can be avoided.

neurots/preprocess/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+
+def preprocess_inputs(params, distrs):
+    """Validate and preprocess all inputs."""
+    validate_neuron_params(params)
+    validate_neuron_distribs(distrs)
+
+    for grow_type in params["grow_types"]:
+        growth_method = params[grow_type]["growth_method"]
+        for preprocess_func in _preprocess_functions[growth_method]:
+            preprocess_func(params[grow_type], distrs[grow_type])
```
eleftherioszisis, line 56: For better maintenance, I would suggest that all preprocessing functions should return modified copies of params and distrs. Or at least copy them once and apply all the modifications to the copies.

tests/test_preprocess.py
```diff
@@ -0,0 +1,79 @@
...
+
+import pytest
+
+from neurots import preprocess
+from neurots.utils import convert_from_legacy_neurite_type
+
+DATA_PATH = Path(__file__).parent / "data"
+
+
+@pytest.mark.parametrize(
```
eleftherioszisis, line 13: I understand that this is convenient, but boy is it hard to debug when sage hits the fan, when you have all these combinations.

eleftherioszisis:
+1 for the design.

neurots/preprocess/__init__.py
adrien-berchet: Yeah, it was just a draft but I agree I will reorganize this.

neurots/preprocess/__init__.py
adrien-berchet: You really like functional programming, don't you? :stuck_out_tongue_winking_eye: 
I am not sure it is relevant to do it each time, because some functions are called each time a grower is instantiated, which will lead to many copies. I think we should copy it only once at the beginning of the grower (which is already done for params but not for distrs).

eleftherioszisis: Sure, that's fine too.

neurots/preprocess/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def inner(func):
+        for i in growth_methods:
+            _preprocess_functions[i].add(func)
+        return func
+
+    return inner
+
+
+@register_preprocess("trunk")
```
lidakanari, line 22: I like this design, where input data are checked depending on the need. I'll get back to this once you integrate the rest of the comments.

neurots/preprocess/__init__.py
lidakanari: I agree with Adrien, if we don't need copies we better not create them...

neurots/generate/algorithms/tmdgrower.py
```diff
@@ -67,15 +67,19 @@ def __init__(
...
-            if stepSZ >= barSZ:
-                L.warning(
-                    "Selected step size %f is too big for bars of size %f",
-                    stepSZ,
-                    barSZ,
-                )
+            self.check_min_bar_length(params, input_data)
+
+    @staticmethod
+    def check_min_bar_length(params, distrs):
```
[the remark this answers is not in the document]
eleftherioszisis: Hmm, if the checks need more info than the parames and distrs, you could have the entire grower passing to each check function to ensure that for all check cases you will always have all the info available.

Having the check attached to the grower however makes it unavailable to use it from another grower for example that may not derive from this one. And it would be strange to import this grower just to use its check statics.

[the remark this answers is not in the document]
adrien-berchet: Yeah ok, it's probably better to get rid of the inheritance limitations indeed. I moved this.

adrien-berchet:
I reorganized the code so that all the check/preprocess functions are in the preprocess module. I created 3 kind of functions:
* validity checkers: they check that the input parameters and distributions given by the user are valid (if they are not the code will probably break or have undefined behavior).
* preprocesses: they can consider that the inputs are already valid ; they can update/modify the parameters and distributions ; it is responsible to ensure the updated parameters and distributions are still valid. (I think this is what @arnaudon needs for the fit in #49 )
* relevancy checkers: they check that the current parameters and distributions will give 'relevant' results ; they can take optional `start_point` and a `context` arguments to refine the checks.

WDYT @eleftherioszisis @lidakanari ?

neurots/generate/algorithms/tmdgrower.py
```diff
@@ -58,24 +59,14 @@ def __init__(
         """TMD basic grower."""
         super().__init__(input_data, params, start_point, context)
         self.bif_method = bif_methods[params["branching_method"]]
-        self.params = copy.deepcopy(params)
```
lidakanari, line 61: I was keeping this for safety reasons, could we keep it even though a copy might not be necessary?

neurots/preprocess/utils.py
```diff
@@ -0,0 +1,68 @@
...
+        for i in growth_methods:
+            _PREPROCESS_FUNCTIONS["validator"][i].add(func)
+        return func
+
+    return inner
+
+
+def preprocess_inputs(params, distrs):
+    """Validate and preprocess all inputs."""
+    params = deepcopy(params)
```
lidakanari, line 55: I see the deepcopy happens here, but what about the case where we do not call the preprocess tools? Is it certain that the inputs are not modified at all when there is no preprocessing?

neurots/schemas/parameters.json
```diff
@@ -145,16 +145,20 @@
                             "properties": {
                                 "mean": {
                                     "description": "The mean of the distribution",
+                                    "minimum": 0,
```
lidakanari, line 148: Good point!

lidakanari (commented):
I think it looks good imo. I let Eleftherios comment for the technical details.

neurots/generate/algorithms/tmdgrower.py
adrien-berchet: I removed it from here because it is already in the `__init__` of `AbstractAlgo`, so calling `super().__init__()` already performs the deep copy.

arnaudon:
LGTM, feel free to merge and update mine if it needs to be adapted to this

neurots/preprocess/utils.py
```diff
@@ -0,0 +1,68 @@
...
+from collections import defaultdict
+from copy import deepcopy
+from itertools import chain
+
+from neurots.validator import validate_neuron_distribs
+from neurots.validator import validate_neuron_params
+
+_PREPROCESS_FUNCTIONS = {
+    "preprocess": defaultdict(set),
+    "validator": defaultdict(set),
```
eleftherioszisis, line 27: Minor: Could we have both keys as verbs, i.e. "preprocess" and "validate" for consistency?

neurots/preprocess/utils.py
```diff
@@ -0,0 +1,68 @@
...
+}
+
+
+def register_preprocess(*growth_methods):
```
eleftherioszisis, line 31: Why not pass a list?

neurots/preprocess/validity_checkers.py
```diff
@@ -0,0 +1,54 @@
...
+from neurots.preprocess.relevancy_checkers import check_min_bar_length
+from neurots.preprocess.utils import register_validator
+
+
+@register_validator("trunk")
+def check_num_seg(params, distrs):
+    """Check that params contains a 'num_seg' entry."""
+    # pylint: disable=unused-argument
+    if "num_seg" not in params:
+        raise KeyError(
```
eleftherioszisis, line 35: Minor: Could we have dedicated errors in the NeuroTS context, so that they can be filtered?

eleftherioszisis:
I have a few minor comments, but overall it's legitimate. One last thing, can we please use "relevance" instead of "relevancy"? I know they are synonyms, but my brain is triggered because "relevance" is the most used word. :D

neurots/preprocess/utils.py
adrien-berchet: I changed to `preprocessors` and `validators` because the stored objects are functions. Is it ok for you?

neurots/preprocess/utils.py
adrien-berchet: So the decorator is easier to write:
```python
@register_preprocess("a", "b", "c")
```
instead of:
```python
@register_preprocess(["a", "b", "c"])
```

neurots/preprocess/validity_checkers.py
adrien-berchet: Sure, I added a specific exception

adrien-berchet:
> I have a few minor comments, but overall it's legitimate. One last thing, can we please use "relevance" instead of "relevancy"? I know they are synonyms, but my brain is triggered because "relevance" is the most used word. :D

I changed for `relevance`, I didn't know it was more used than `relevancy`, I guess I'm an old school guy :)

neurots/preprocess/utils.py
eleftherioszisis: Yeah, as far as they are consistent. :+1:

neurots/preprocess/utils.py
eleftherioszisis: It's fine either way. I will resolve this.

eleftherioszisis:
I am good with the changes. @lidakanari I leave it to you to approve.

arnaudon:
I just rebased on main, I guess if CI pass we can merge?

arnaudon (approved):

PR #54, opened 2023-01-27
Fix ci

adrien-berchet:


arnaudon:
be careful, this is based on my diameter branch I think, don't merge on main

adrien-berchet:
> be careful, this is based on my diameter branch I think, don't merge on main

Yeah it's just for testing. I will port the updates to your branch if I can make it work here.

adrien-berchet:
This was ported to #47

PR #61, opened 2023-04-20
Add context handling with accept/reject

arnaudon:
This PR adds an accept/reject mechanisms to allow for external modification of section and trunk growths via the generic `context` object.

### DIFF
--- a/neurots/generate/section.py
+++ b/neurots/generate/section.py
@@ -102,7 +102,7 @@ class SectionGrower:
     def next_point(self, current_point):
         """Returns the next point depending on the growth method and the previous point.
 
-        If a context is present, an accept-reject mechanisms will be used to alter the next point.
+        If a context is present, an accept-reject mechanism will be used to alter the next point.
         """
         if self.context is not None and "constraints" in self.context:  # pragma: no cover
 
--- a/neurots/generate/section.py
+++ b/neurots/generate/section.py
@@ -19,6 +19,10 @@ DISTANCE_MIN = 1e-8
 # Memory decreases with distance from current point
 WEIGHTS = np.exp(np.arange(1, MEMORY + 1) - MEMORY)
 
+# default parameters for accept/reject
+DEFAULT_MAX_TRIES = 100
+DEFAULT_NOISE_INCREASE = 0.5
+
 
 class SectionGrower:
     """Class for the section growth.
@@ -81,45 +85,43 @@ class SectionGrower:
         """Increases the path distance."""
         self.pathlength += length
 
-    def next_point(self, current_point):
-        """Returns the next point depending on the growth method and the previous point."""
+    def _propose(self, noise=0.0):
+        """Propose the direction for a next section point.
 
-        def prob(data):
-            """Probability function to accept a next point.
-
-            It will always accept unless a context prob function is present.
-            """
-            if self.context is not None and "section_prob" in self.context:  # pragma: no cover
-                return self.context["section_prob"](data, current_point=current_point)
-            return 1.0
-
-        def propose(noise=0.0):
-            """Propose a next section point.
-
-            Args:
-                noise (float): artificially increase the randomness to allow for context changes
-            """
-            direction = (
-                self.params.targeting * self.direction
-                + self.params.randomness
-                * get_random_point(random_generator=self._rng)
-                * (1 + noise)
-                + self.params.history * self.history()
-            )
-
-            return direction / vectorial_norm(direction)
-
-        direction = accept_reject(
-            propose,
-            prob,
-            self._rng,
-            max_tries=self.context.get("params_section", {}).get("max_tries", 100)
-            if self.context is not None
-            else 100,
-            noise_increase=self.context.get("params_section", {}).get("noise_increase", 0.5)
-            if self.context is not None
-            else 0.5,
+        Args:
+            noise (float): artificially increase the randomness to allow for context changes
+        """
+        direction = (
+            self.params.targeting * self.direction
+            + self.params.randomness * get_random_point(random_generator=self._rng) * (1.0 + noise)
+            + self.params.history * self.history()
         )
+
+        return direction / vectorial_norm(direction)
+
+    def next_point(self, current_point):
+        """Returns the next point depending on the growth method and the previous point.
+
+        If a context is present, an accept-reject mechanisms will be used to alter the next point.
+        """
+        if self.context is not None and "section_prob" in self.context:  # pragma: no cover
+            direction = accept_reject(
+                self._propose,
+                self.context["section_prob"],
+                self._rng,
+                max_tries=self.context.get("params_section", {}).get("max_tries", DEFAULT_MAX_TRIES)
+                if self.context is not None
+                else DEFAULT_MAX_TRIES,
+                noise_increase=self.context.get("params_section", {}).get(
+                    "noise_increase", DEFAULT_NOISE_INCREASE
+                )
+                if self.context is not None
+                else DEFAULT_NOISE_INCREASE,
+                current_point=current_point,
+            )
+        else:
+            direction = self._propose()
+
         seg_length = self.step_size_distribution.draw_positive()
         next_point = current_point + seg_length * direction
         self.update_pathlength(seg_length)
--- a/neurots/generate/section.py
+++ b/neurots/generate/section.py
@@ -104,7 +104,7 @@ class SectionGrower:
 
         If a context is present, an accept-reject mechanism will be used to alter the next point.
         """
-        if self.context is not None and "constraints" in self.context:  # pragma: no cover
+        if self.context is not None and self.context.get("constraints", []):  # pragma: no cover
 
             def prob(*args, **kwargs):
                 p = 1.0
--- a/neurots/generate/section.py
+++ b/neurots/generate/section.py
@@ -108,7 +108,7 @@ class SectionGrower:
 
             def prob(*args, **kwargs):
                 p = 1.0
-                for i, constraint in enumerate(self.context["constraints"]):
+                for constraint in self.context["constraints"]:
                     p *= constraint["section_prob"](*args, **kwargs)
                 return p
 
--- a/neurots/generate/section.py
+++ b/neurots/generate/section.py
@@ -84,7 +84,7 @@ class SectionGrower:
     def next_point(self, current_point):
         """Returns the next point depending on the growth method and the previous point."""
 
-        def prob(proposal):
+        def prob(_):
             return 1.0
 
         def propose():
--- a/neurots/generate/section.py
+++ b/neurots/generate/section.py
@@ -84,10 +84,17 @@ class SectionGrower:
     def next_point(self, current_point):
         """Returns the next point depending on the growth method and the previous point."""
 
-        def prob(_):
+        def prob(data):
+            """Probability function to accept a next point.
+
+            It will always accept unless a context prob function is present.
+            """
+            if "section_prob" in self.context:  # pragma: no cover
+                return self.context["section_prob"](data)
             return 1.0
 
         def propose():
+            """Propose a next section point."""
             direction = (
                 self.params.targeting * self.direction
                 + self.params.randomness * get_random_point(random_generator=self._rng)
--- a/neurots/generate/section.py
+++ b/neurots/generate/section.py
@@ -89,7 +89,7 @@ class SectionGrower:
         """Propose the direction for a next section point.
 
         Args:
-            extra_randomness (float): artificially increase the randomness to allow for context changes
+            extra_randomness (float): artificially increase the randomness to allow for context
         """
         direction = self.params.targeting * self.direction + self.params.history * self.history()
         if extra_randomness > -1:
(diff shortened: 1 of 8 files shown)


arnaudon:
@adrien-berchet , when you have 5min, if you can take a look at this to see if the general implementation makes sense. I will then use context to modify the prob() function.

neurots/utils.py
```diff
@@ -106,3 +106,26 @@ def point_to_section_segment(neuron, point, rtol=1e-05, atol=1e-08):
...
+        if _prob == 1.0:
+            # this ensures we don't change rng for the tests, but its not really needed
+            return proposal
+
+        if rng.binomial(1, _prob):
+            return proposal
+        n_try += 1
+    warnings.warn(
+        """We could not sample from distribution, so we take a random point.
+                    Consider checking the given probability distribution."""
```
adrien-berchet, line 127: Why a new line here?

tox.ini
```diff
@@ -61,7 +61,7 @@ commands =
 basepython = python3.8
 
 [testenv:lint]
-basepython = python3.8
+basepython = python3
```
adrien-berchet, line 64: I see you here too :stuck_out_tongue_winking_eye:

tests/test_utils.py
```diff
@@ -113,3 +113,34 @@ def test_point_to_section_segment():
...
+    for _ in range(10):
+        val = utils.accept_reject(propose, prob, rng, null=null)
+        assert val == 1.0
+
+    def propose_null():
+        return 0
+
+    # check if we attain max_tries we return null = 0
+    val = utils.accept_reject(propose_null, prob, rng, null=null)
+    assert val == 0.0
```
adrien-berchet, line 142: Maybe it would be more demonstrating if the value returned by `null` a completely different, like `-1` or `None` or something like this.

adrien-berchet:
> @adrien-berchet , when you have 5min, if you can take a look at this to see if the general implementation makes sense. I will then use context to modify the prob() function.

Looks nice!

.pre-commit-config.yaml
```diff
@@ -1,5 +1,5 @@
 default_language_version:
-  python: python3.8
+  python: python3
```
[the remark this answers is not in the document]
arnaudon: haha, it's temporary, I'll revert later ;)

tox.ini
arnaudon: ta ta ta

tests/test_utils.py
arnaudon: 0 is different from 1, no? :P

.pre-commit-config.yaml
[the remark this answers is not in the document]
adrien-berchet: 😜

neurots/utils.py
adrien-berchet, line 127:
```suggestion
        "We could not sample from distribution, so we take a random point. "
        "Consider checking the given probability distribution."
```
Just this?

tests/test_utils.py
adrien-berchet: Sure, but `0` can come either from the `propose` or the `null` function. Having a value specific to the `null` function would be more clear IMO.

arnaudon: agreed, done

tests/test_utils.py
```diff
@@ -113,3 +113,34 @@ def test_point_to_section_segment():
...
+
+    # check we always return 1
+    for _ in range(10):
+        val = utils.accept_reject(propose, prob, rng, null=null)
+        assert val == 1.0
+
+    def propose_null():
+        return 0.0
+
+    # check if we attain max_tries we return null = 0
```
adrien-berchet, line 140:
```suggestion
    # check if we attain max_tries we return null = -1
```

neurots/generate/section.py
```diff
@@ -91,15 +96,43 @@ def update_pathlength(self, length):
...
+
+        If a context is present, an accept-reject mechanisms will be used to alter the next point.
+        """
+        if self.context is not None and "section_prob" in self.context:  # pragma: no cover
+            direction = accept_reject(
+                self._propose,
+                self.context["section_prob"],
+                self._rng,
+                max_tries=self.context.get("params_section", {}).get("max_tries", DEFAULT_MAX_TRIES)
+                if self.context is not None
```
adrien-berchet, line 124: We already know that `self.context` is not None here since we tested it at line 118

neurots/generate/section.py
```diff
@@ -91,15 +96,43 @@ def update_pathlength(self, length):
...
+                else DEFAULT_MAX_TRIES,
+                noise_increase=self.context.get("params_section", {}).get(
+                    "noise_increase", DEFAULT_NOISE_INCREASE
+                )
+                if self.context is not None
```
adrien-berchet, line 129: Same as L124

neurots/generate/section.py
arnaudon: good catch, I simplified things and forgot that one

arnaudon:
@lidakanari @adrien-berchet , this can be reviewed as well, again, it is experimental, so not fully tested

neurots/generate/section.py
```diff
@@ -91,15 +96,58 @@ def update_pathlength(self, length):
...
             + self.params.history * self.history()
         )
 
-        direction = direction / vectorial_norm(direction)
+        return direction / vectorial_norm(direction)
+
+    def next_point(self, current_point):
+        """Returns the next point depending on the growth method and the previous point.
+
+        If a context is present, an accept-reject mechanisms will be used to alter the next point.
```
adrien-berchet, line 116:
```suggestion
        If a context is present, an accept-reject mechanism will be used to alter the next point.
```

neurots/generate/tree.py
```diff
@@ -172,6 +172,20 @@ def add_section(
...
+                for constraint in self.context["constraints"]
+                if "section_prob" in constraint
+                and SectionType(self.params["tree_type"]).name
+                in constraint.get("neurite_types", [])
+                and process in constraint.get("processes", ["major", "secondary"])
+            ]
+            if constraints:
+                context["constraints"] = constraints
+            else:
+                del context["constraints"]
```
adrien-berchet, line 188: Why do you prefer to delete it instead of keeping an empty list?

neurots/generate/tree.py
```diff
@@ -248,8 +262,21 @@ def next_point(self):
...
+                # the original topology
+                if section_grower.process == "major":
+                    _term = section_grower.stop_criteria["TMD"].term
+                    section_grower.stop_criteria["TMD"].term = self.params["cut_major_length"]
+
             state = self.growth_algo.extend(section_grower)
 
+            if "cut_major_length" in self.params:  # pragma: no cover
+                if section_grower.process == "major":
+                    section_grower.stop_criteria["TMD"].term = _term
```
adrien-berchet, line 278: This is already in #78

neurots/utils.py
```diff
@@ -106,3 +106,27 @@ def point_to_section_segment(neuron, point, rtol=1e-05, atol=1e-08):
             return section.id, offset[0][0]
 
     raise ValueError(f"Cannot find point in morphology that matches: {point}")
+
+
+def accept_reject(
+    propose, probability, rng, null=None, max_tries=100, noise_increase=0.5, **probability_kwargs
+):
+    """Generic accept reject algorithm."""
```
adrien-berchet, line 114:
```suggestion
    """Generic accept/reject algorithm."""
```

neurots/generate/tree.py
adrien-berchet: ah ok, so #78 should be merge before this one and then you'll rebase?

neurots/generate/tree.py
adrien-berchet: Ok but I feel it dangerous because the result of `SectionGrower.next_point()` will be different whether `self.context["constraints"]` does not exist or if it exists with an empty list for example. This means that you suppose that `SectionGrower.next_point()` is always called after `TreeGrower.add_section()`, but this is actually not mandatory.
Maybe it would be safer to do something like `if context.get("constraints", [])` in `SectionGrower.next_point()`? And so you can leave an empty list in `TreeGrower.add_section()`.

neurots/generate/tree.py
arnaudon: yes, so i can use both together

neurots/generate/tree.py
adrien-berchet: I think it's safer :)

arnaudon: done!

neurots/utils.py
```diff
@@ -106,3 +106,27 @@ def point_to_section_segment(neuron, point, rtol=1e-05, atol=1e-08):
...
+        _prob = probability(proposal, **probability_kwargs)
+        if _prob == 1.0:
+            # this ensures we don't change rng for the tests, but its not really needed
+            return proposal
+
+        if rng.binomial(1, _prob):
+            return proposal
+        n_tries += 1
+    warnings.warn(
+        "We could not sample from distribution, so we take a random point. "
```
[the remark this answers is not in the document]
arnaudon: done!

eleftherioszisis:
I am concerned about this PR. There are a lot of changes affecting the core functionality in many places and almost no tests to test the new functionality thoroughly. In addition, there are no tests for the expected behavior of the algorithm that should validate the sampling algorithm behavior.

By just looking at the review, I doubt any of us can guarantee that these changes are not breaking or introducing unexpected behavior.

And on a different note, is there a performance benchmark for this new approach? accept/reject is usually quite slower and synthesis already takes quite some time when considered from a large-scale point of view. Not a priority, I know, but a ballpark estimation to make sure we don't introduce a 100x slowdown would be nice.

arnaudon:
I'm not sure which 'core functionalities' you are mentioning? its all under ```if self.context is not None``` no?

adrien-berchet:
What about moving this feature in a mixin so it would not change anything on the usual stuff and it's still possible to use it for external applications?

eleftherioszisis:
> I'm not sure which 'core functionalities' you are mentioning? its all under `if self.context is not None` no?

I see changes in `section.py`, `tree.py`, `orientations.py`. All these are core modules.

Also, other parts of the code use a context too. For instance:
https://github.com/BlueBrain/NeuroTS/blob/9e586f72e23fe7d936571f1f97ef47f15c88338e/neurots/astrocyte/context.py#L54

Maybe it doesn't matter, but keep in mind that the generic classes should not cover specific use cases encountered only in neurons. This is what derived classes are for.

The used context should have an interface rather than accessing keys at random in different parts of the code so that users/devs can understand what they are expected to pass. This can be as simple as having a json schema or a class that represents that context, whichever is preferable. Having an undefined dictionary that could have anything in it, makes the code very hard to use.

> What about moving this feature in a mixin so it would not change anything on the usual stuff and it's still possible to use it for external applications?

Not sure. If you wish we could discuss it in more detail.

arnaudon:
What would be your suggestion to improve then? I don't mind leaving this open for ever, it does the job I need it to do with the other open PR on region-grower, which is almost over with bbp end. I can surely add a schema here, which I would have to overwrite in region-grower (as the other schemas) so I can have more freedom from that side. This is supposed to be a 'hidden' feature, not a core feature, but it enters it at the 'core' of the code, its not my fault ;) If I recall coding this 3y ago or so, i tried to be minimalistic in the implementation, but I'm open to suggestions.

eleftherioszisis:
> What would be your suggestion to improve then? I don't mind leaving this open for ever, it does the job I need it to do with the other open PR on region-grower, which is almost over with bbp end. I can surely add a schema here, which I would have to overwrite in region-grower (as the other schemas) so I can have more freedom from that side. This is supposed to be a 'hidden' feature, not a core feature, but it enters it at the 'core' of the code, its not my fault ;) If I recall coding this 3y ago or so, i tried to be minimalistic in the implementation, but I'm open to suggestions.

I would say mainly test that all the new parts of the code work as expected and produce the expected sampling.

And yeah, having a schema for the context will help make things standardized. Why would it be a hidden feature though?

arnaudon:
I can try to add a couple more tests, but this code is not meant to be use 'directly'. There is no probability function implemented here, it has to be passed by a user externally via the context. The 'success' of the sampling will very much depend on this proba function, as you know. It is hidden in this sense: its not easy to use at all, and very experimental in term of writing a good probability function to do what the user needs. In this case, the user is me only, via region-grower (not OSS). Also, hidden in the sense: unless you know the specific way to access this part of the code (via me, or region-grower), you will not find it ;)

arnaudon:
@adrien-berchet , I'll look at you suggestion soon, clean up a few things and check the tests maybe later this week

adrien-berchet:
And what about just improving how external applications can customize the tree and section growers of NeuroTS? So we could just define a new custom grower in `region-grower` with a specific behavior and pass it to NeuroTS.

eleftherioszisis:
> And what about just improving how external applications can customize the tree and section growers of NeuroTS? So we could just define a new custom grower in `region-grower` with a specific behavior and pass it to NeuroTS.

Why not specify that custom grower here? It would be easier to ensure with tests that changes in NeuroTS are compatible with the custom growers.

adrien-berchet:
> Why not specify that custom grower here? It would be easier to ensure with tests that changes in NeuroTS are compatible with the custom growers.
> 
> This is what I did for astrocytes for example.

It would be just to make it clear that this feature is specific to region-grower, not a core feature. For example in axon-synthesis I defined another custom grower (with another method to handle boundaries).

But anyway, when no context is given this PR only adds 2 `if` statements so it should not have any impact on the performance. So I think that we should mainly decide if this feature should be part of the core features or just optional and focus on improving the test coverage and the readability of the code.

arnaudon:
tests are fine locally, weird weird

adrien-berchet (approved):

PR #68, opened 2023-05-17
Fix: Relax soma/first point check

arnaudon:
If one has a soma located far from origin, the rtol may consider the point to be too close, hence its better to only use atol.

### DIFF
--- a/neurots/generate/soma.py
+++ b/neurots/generate/soma.py
@@ -87,7 +87,7 @@ class Soma:
 
     def orientation_from_point(self, point):
         """Return the orientation from the soma center to a point on the soma surface."""
-        if np.allclose(point, self.center, atol=0.01, rtol=0):
+        if np.allclose(point, self.center, rtol=0):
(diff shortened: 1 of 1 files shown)


neurots/generate/soma.py
```diff
@@ -87,7 +87,7 @@ def point_from_trunk_direction(self, phi, theta):
 
     def orientation_from_point(self, point):
         """Return the orientation from the soma center to a point on the soma surface."""
-        if np.allclose(point, self.center):
+        if np.allclose(point, self.center, atol=0.01, rtol=0):
```
adrien-berchet, line 90: Why `atol=0.01`?

arnaudon, line 90:
```suggestion
        if np.allclose(point, self.center, rtol=0):
```

adrien-berchet: :joy:

adrien-berchet (approved):

PR #70, opened 2023-06-20
Fix: Allow for normal_pia to have any pia direction

arnaudon:


arnaudon:
> Shouldn't the `pia_direction` parameter be added to the schema?

It's use 'on the fly', its not really to write in the json file atm, but we can add it to schema yes

adrien-berchet:
There is actually a missing `"additionalProperties": false` in the last element of `definitions/neurite/orientation/values` :scream: 
I guess that's why it works?

adrien-berchet (approved):

PR #72, opened 2023-06-22
Build: Relax numpy max version

adrien-berchet:


arnaudon (approved):
all good, it works on bb5 as well, thank you!

PR #75, opened 2023-08-28
Fix: Handle near symetrical two major apical branches

arnaudon:
Some explanation:
- to fix the asymetrical issue, I realised that the pathlength of the two branches was not identical, and the one with bif was larger, hence it will get more bif than the other. 
- I first thought it was because of random seg_length, so I freeze to 1
- then I realised that it was also (or mostly) due to the fact that a .fist_point is called after a big, hence it had 1 micron to pathlength when a bif occurs, hence this branch is more likely to get the next bif
- I also tried a new argument to have 2 major branches, and control there angle, so it looks more like the hipp reconstructions

### DIFF
--- a/neurots/generate/tree.py
+++ b/neurots/generate/tree.py
@@ -120,6 +120,7 @@ class TreeGrower:
         self.active_sections = []
         self.context = context
         self._rng = random_generator
+        self.two_major = 1  # local hack to have two major branches in first bif
 
         # Creates the distribution from which the segment lengths
         # To sample a new seg_len call self.seg_len.draw()
@@ -259,14 +260,17 @@ class TreeGrower:
                     section_grower.id = section.id
                     # the current section_grower bifurcates
                     # Returns two section_grower dictionaries: (S1, S2)
-                    for child_section in self.growth_algo.bifurcate(section_grower):
+                    for child_section in self.growth_algo.bifurcate(
+                        section_grower, two_major=self.two_major
+                    ):
+                        self.two_major = 0
                         child = self.add_section(
                             parent=section, pathlength=section_grower.pathlength, **child_section
                         )
                         # Copy the final normed direction of parent to all children
                         child.latest_directions.append(latest)
                         # Generate the first point of the section
-                        child.first_point()
+                        # child.first_point()
                     self.active_sections.remove(section_grower)
 
                 elif state == "terminate":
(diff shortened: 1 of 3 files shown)


neurots/extract_input/input_parameters.py
```diff
@@ -74,7 +74,7 @@ def merged_params(data):
                 "growth_method": method,
                 "branching_method": branching,
                 "modify": None,
-                "step_size": {"norm": {"mean": 1.0, "std": 0.2}},
```
arnaudon, line 77: if we have noise here, there will be some imbalance. I am not sure why we would need random step_size, I guess uniform one with 1 micron is also ok, no? If not, we can either leave original value here, but we need to fix to 0 for hipp cells. I would be more in favor of changing default, in case other users bumped into the same issue.

neurots/generate/tree.py
```diff
@@ -120,6 +120,7 @@ def __init__(
         self.active_sections = []
         self.context = context
         self._rng = random_generator
+        self.two_major = parameters.get("two_major", False)
```
arnaudon: this is to have two sub-apical tree that bifurcate from the first bif after soma, and both go up. Without this, one will be a trunk, and the other will go on the side as an oblique. This is specific to the hipp PC cells, but may be found elsewhere?

neurots/generate/tree.py
```diff
@@ -277,3 +283,6 @@ def next_point(self):
                 # TODO: Can the state be something else than "bifurcate" or "terminate" here?
                 # If not we could add an else statement to raise a warning or an exception if the
                 # value is anything else.
+            elif state == "continue":
```
arnaudon, line 286: this is the main issue for imbalance: we add a child.first_point() at each bif, which makes the pathlength longer for branches with bifs, hence more likely to get more bifs later. We can either remove child.first_point(), or add a .next() to match the pathlengths.

arnaudon:
Many test break due to adding one more point in grower, if we agree on this fix, I'll update them

neurots/generate/tree.py
KeremKurban: Sometimes its not 2 major but 3 or even 4 depending on the species. 
image
and neocortex also have these morphologies.

neurots/generate/tree.py
```diff
@@ -277,3 +277,6 @@ def next_point(self):
...
```
adrien-berchet, line 280:
```suggestion
            else:
```
Isn't it enough?

Issue #76, opened 2023-08-30
[Bug report] Soma type needs to be set for new MorphIO


arnaudon:
### Describe the bug

with new morphio, we get:

```Soma must be a contour for ASC and H5: see https://github.com/BlueBrain/MorphIO/issues/457```

### To Reproduce

```Python
make a neuron and save in .asc
```

### Error

```Python traceback
Soma must be a contour for ASC and H5: see https://github.com/BlueBrain/MorphIO/issues/457
```

### neurots Version in Use

latest

### Python Version

any

### Operating system

Linux

arnaudon:
solved in #77

PR #77, opened 2023-08-31
Fix: explicitly write soma type

arnaudon:
solved #76


arnaudon:
I don't get why the try is not covered, as it crashed here: https://github.com/BlueBrain/NeuroTS/actions/runs/6037160854/job/16380883990

adrien-berchet (approved):

PR #78, opened 2023-08-31
Feat: Add major_termination_length parameter

arnaudon:
This allows one to create an axon with a main branch and collaterals. The main branch is usually cut for local axons, as it propagates to other regions. With this PR, we can have a very long bar for it (so that it is considered as major branch), but not synthesize it entirely, to mimic the local axon reconstructions.

### DIFF
--- a/neurots/generate/tree.py
+++ b/neurots/generate/tree.py
@@ -248,7 +248,20 @@ class TreeGrower:
         for section_grower in ordered_sections:
             # the current section_grower is generated
             # In here the stop criterion can be modified accordingly
+
+            if "cut_major_length" in self.params:
+                # this makes an early termination of major branch
+                # it needs to revert the value after getting the state to preserve
+                # the original topology
+                if section_grower.process == "major":
+                    _term = section_grower.stop_criteria["TMD"].term
+                    section_grower.stop_criteria["TMD"].term = self.params["cut_major_length"]
+
             state = self.growth_algo.extend(section_grower)
+
+            if "cut_major_length" in self.params:
+                if section_grower.process == "major":
+                    section_grower.stop_criteria["TMD"].term = _term
 
             if state != "continue":
                 section = self.append_section(section_grower)
--- a/neurots/generate/tree.py
+++ b/neurots/generate/tree.py
@@ -251,7 +251,7 @@ class TreeGrower:
 
             if (
                 section_grower.process == "major" and "major_termination_length" in self.params
-            ):  # pragma: no cover
+            ):
                 # this makes an early termination of major branch
                 # it needs to revert the value after getting the state to preserve
                 # the original topology
@@ -262,7 +262,7 @@ class TreeGrower:
 
             state = self.growth_algo.extend(section_grower)
 
-            if _term is not None:  # pragma: no cover
+            if _term is not None:
                 section_grower.stop_criteria["TMD"].term = _term
 
             if state != "continue":
--- a/neurots/generate/tree.py
+++ b/neurots/generate/tree.py
@@ -249,7 +249,9 @@ class TreeGrower:
             # the current section_grower is generated
             # In here the stop criterion can be modified accordingly
 
-            if section_grower.process == "major" and "cut_major_length" in self.params:  # pragma: no cover
+            if (
+                section_grower.process == "major" and "cut_major_length" in self.params
+            ):  # pragma: no cover
                 # this makes an early termination of major branch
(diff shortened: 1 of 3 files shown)


arnaudon:
@lidakanari @adrien-berchet , this is ready to review, it is a 'minor hidden thing' for hipp axons, so i didn't add tests for now

neurots/generate/tree.py
```diff
@@ -248,8 +248,21 @@ def next_point(self):
...
+                # the original topology
+                if section_grower.process == "major":
+                    _term = section_grower.stop_criteria["TMD"].term
+                    section_grower.stop_criteria["TMD"].term = self.params["cut_major_length"]
+
             state = self.growth_algo.extend(section_grower)
 
+            if "cut_major_length" in self.params:  # pragma: no cover
+                if section_grower.process == "major":
+                    section_grower.stop_criteria["TMD"].term = _term
```
adrien-berchet, line 264:
```suggestion
            if section_grower.process == "major" and "cut_major_length" in self.params:
                # this makes an early termination of major branch
                # it needs to revert the value after getting the state to preserve
                # the original topology
                _term = section_grower.stop_criteria["TMD"].term
                section_grower.stop_criteria["TMD"].term = self.params["cut_major_length"]

            state = self.growth_algo.extend(section_grower)

            if section_grower.process == "major" and "cut_major_length" in self.params:
                section_grower.stop_criteria["TMD"].term = _term
```
What about this?

neurots/generate/tree.py
```diff
@@ -248,8 +248,19 @@ def next_point(self):
...
+                # this makes an early termination of major branch
+                # it needs to revert the value after getting the state to preserve
+                # the original topology
+                _term = section_grower.stop_criteria["TMD"].term
+                section_grower.stop_criteria["TMD"].term = self.params["cut_major_length"]
+
             state = self.growth_algo.extend(section_grower)
 
+            if section_grower.process == "major" and "cut_major_length" in self.params:
+                section_grower.stop_criteria["TMD"].term = _term
```
adrien-berchet, line 262:
```suggestion
            if section_grower.process == "major" and "cut_major_length" in self.params:  # pragma: no cover
                # this makes an early termination of major branch
                # it needs to revert the value after getting the state to preserve
                # the original topology
                _term = section_grower.stop_criteria["TMD"].term
                section_grower.stop_criteria["TMD"].term = self.params["cut_major_length"]
            else:
                _term = None

            state = self.growth_algo.extend(section_grower)

            if _term is not None:  # pragma: no cover
                section_grower.stop_criteria["TMD"].term = _term
```
Another idea to improve perfs a bit (avoid duplicated checks)

adrien-berchet (commented):
Do you plan to add test in this PR?

arnaudon:
> Do you plan to add tests in this PR?

nope

neurots/generate/tree.py
```diff
@@ -248,8 +248,23 @@ def next_point(self):
...
+                # it needs to revert the value after getting the state to preserve
+                # the original topology
+                _term = section_grower.stop_criteria["TMD"].term
+                section_grower.stop_criteria["TMD"].term = self.params["cut_major_length"]
+            else:
+                _term = None
+
             state = self.growth_algo.extend(section_grower)
 
+            if _term is not None:  # pragma: no cover
```
lidakanari, line 265: I'm not sure I understand this hack. Why not add it to the if / else statement?

neurots/generate/tree.py
```diff
@@ -248,8 +248,23 @@ def next_point(self):
         for section_grower in ordered_sections:
             # the current section_grower is generated
             # In here the stop criterion can be modified accordingly
+
+            if (
+                section_grower.process == "major" and "cut_major_length" in self.params
```
lidakanari, line 253: I would give a more informative name to this parameter: 
"early_terminate_major"
"major_termination_length"

arnaudon: ok, for `major_termination_length'!

neurots/generate/tree.py
arnaudon: because the previous line is outside the if. The second if is to revert the .term change, so that the algo is not affected, but only the .extend function

arnaudon: @lidakanari , let me know if you think of a better implementation, otherwise we can merge and move on to the other PR.

adrien-berchet (approved):
LGTM!
@lidakanari are you also ok?

lidakanari (approved):

PR #85, opened 2023-09-20
Feat: Add a uniform diameter model and deprecate the 'radius' parameter

adrien-berchet:
### Description
The 'radius' parameter is used as default diameter value when we add sections but it's supposed to overwritten by the `NeuronGrower._diametrize` method, so we can just use an internal default value. Nevertheless, in some specific cases the diametrizer does nothing (that's actually the case in many tests), which I think is a bit weird. But to reproduce this behavior if someone needs it I added a `uniform` model which can take a diameter value and apply it to the entire neurites.

### Checklist

This pull request is:

- [ ] A documentation / typographical error fix
	- Good to go, no issue or tests are needed
- [ ] A short code fix
	- [ ] Please include: `Fixes: #<issue number>` in the description if it solves an existing issue
	  (which must include a complete example of the issue).
	- [ ] Please include tests that fail with the `main` branch and pass with the provided fix.
- [x] A new feature implementation or update an existing feature
	- [ ] Please include: `Fixes: #<issue number>` in the description if it solves an existing issue
	  (which must include a complete example of the feature).
	- [x] Please include tests that cover every lines of the new/updated feature.
	- [ ] Please update the documentation to describe the new/updated feature.

### DIFF
--- a/neurots/generate/diametrizer.py
+++ b/neurots/generate/diametrizer.py
@@ -355,6 +355,24 @@ def diametrize_constant_per_neurite(neuron, neurite_type=None):
             sec.diameters = mean_diam * np.ones(len(sec.diameters))
 
 
+def diametrize_uniform(neuron, neurite_type=None, *, diam_params):
+    """Set all diameters of a morphio-neuron to a given value.
+
+    Args:
+        neuron (morphio.mut.Morphology): The morphology that will be diametrized.
+        neurite_type (morphio.SectionType): Only the neurites of this type are diametrized.
+        diam_params (dict): The model parameters (should only contain a 'diameter' entry).
+    """
+    diameter = diam_params.get(neurite_type.name, None)
+    if diameter is None:
+        return
+    roots = root_section_filter(neuron, neurite_type)
+
+    for root in roots:
+        for sec in root.iter():
+            sec.diameters = diameter * np.ones(len(sec.diameters))
+
+
 def diametrize_smoothing(neuron, neurite_type=None):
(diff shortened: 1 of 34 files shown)


neurots/generate/diametrizer.py
```diff
@@ -355,6 +355,24 @@ def diametrize_constant_per_neurite(neuron, neurite_type=None):
...
+def diametrize_uniform(neuron, neurite_type=None, *, diam_params):
+    """Set all diameters of a morphio-neuron to a given value.
+
+    Args:
+        neuron (morphio.mut.Morphology): The morphology that will be diametrized.
+        neurite_type (morphio.SectionType): Only the neurites of this type are diametrized.
+        diam_params (dict): The model parameters (should only contain a 'diameter' entry).
+    """
+    diameter = diam_params.get(neurite_type.name, None)
+    if diameter is None:
```
arnaudon, line 367: can you instead put a DEFAULT_DIAMETER=1 ?

adrien-berchet: Sure, done

adrien-berchet: Ah wait, I think I misunderstood. Do you mean instead of doing nothing when it's `None`?

arnaudon: yes, no? I'm not sure either what it does if you hit this `return`

arnaudon: ah ok ok, all good then!

arnaudon (approved):

PR #86, opened 2023-09-21
Fix: Allow for non-default neurite_types

arnaudon:


arnaudon:
@adrien-berchet , can you approve this quick? thanks!

adrien-berchet (approved):

arnaudon:
faster than lucky luke!

PR #89, opened 2023-09-28
Feat: Allow for single basal

arnaudon:
Some cell types have a unique basal, hence we disable this check by setting a parameter to enforce at least 1, not 2.


adrien-berchet (commented):
LGTM, can you just fix the coverage plz?

arnaudon:
> LGTM, can you just fix the coverage plz?

done! (and improved the print message)

adrien-berchet (approved):
Thanks!

PR #90, opened 2023-09-28
Fix: Allow for normal orientation with basals

arnaudon:


arnaudon:
The test values are different, just because of the call to sample.n_neurites which shifts the random numbers.

neurots/generate/orientations.py
```diff
@@ -234,24 +234,30 @@ def _mode_normal_pia_constraint(self, values_dict, _):
...
+                    else:
+                        thetas.append(0)
+                else:
+                    thetas.append(np.clip(self._rng.normal(mean, std), 0, np.pi))
+
+            phis = self._rng.uniform(0, 2 * np.pi, len(means))
+            angles += spherical_angles_to_pia_orientations(
+                phis, thetas, self._parameters.get("pia_direction", None)
+            ).tolist()
+        print(angles)
```
adrien-berchet, line 259:
```suggestion
```

adrien-berchet (changes requested):

adrien-berchet (approved):

PR #95, opened 2023-11-27
Fix: Make bifurcation angles globally invariant

arnaudon:
If one sets a non-[0,1,0] pia_direction, as it is the case for insitu synthesis, bifurcation angles are 'wrong', in the sense that they will depend on the pia_direction. As this is a global rotation of a cell, the bifurcation angles should be invariant to pia_direction.

As an example, here is a synthesized cell on a plane (z=0 always) with rotation invariance (this PR) and pia_direction = [1, 1, 1]: 
Screenshot 2023-11-27 at 13 49 02

and the same setup without this PR: 

Screenshot 2023-11-27 at 13 53 31

we see in the first case the original plane (z=0) rotated to match y-> pia_direction=[1,1,1].

### DIFF
--- a/examples/synthesize_single_neuron_y_direction.py
+++ b/examples/synthesize_single_neuron_y_direction.py
@@ -16,16 +16,17 @@
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
 """
-Synthesize a single neuron
-==========================
+Synthesize a single neuron with global direction
+================================================
 
-This example shows how to synthesize a single cell with simple parameters.
+This example shows how to synthesize a single cell with different y directions
 """
 
 import json
 from pathlib import Path
 
 import numpy as np
+from morph_tool.transform import rotate
 
 import neurots
 
@@ -33,7 +34,8 @@ import neurots
 def run(output_dir, data_dir):
     """Run the example for generating a single cell."""
     np.random.seed(42)
-    params = json.load(open(data_dir / "bio_params.json"))
+    with open(data_dir / "bio_params.json", encoding="utf-8") as p_file:
+        params = json.load(p_file)
     # use trunk angle with y_direction awareness
     params["basal_dendrite"]["orientation"] = {
         "mode": "pia_constraint",
@@ -62,7 +64,7 @@ def run(output_dir, data_dir):
     N = neurots.NeuronGrower(
         input_distributions=data_dir / "bio_distr.json",
         input_parameters=params,
-        #context={"y_direction": [0.0, 1.0, 0.0]},
+        # context={"y_direction": [0.0, 1.0, 0.0]},
     )
 
     # Grow the neuron
@@ -87,7 +89,6 @@ def run(output_dir, data_dir):
     neuron.write(output_dir / "generated_cell_x.asc")
 
     # the rotated neuron should be the same as original one
-    from morph_tool.transform import rotate
 
     rotate(neuron, [[0, -1, 0], [1, 0, 0], [0, 0, 1]])
     neuron.write(output_dir / "generated_cell_x_rot.asc")
--- a/examples/synthesize_single_neuron_y_direction.py
+++ b/examples/synthesize_single_neuron_y_direction.py
@@ -1,19 +1,8 @@
 # noqa
 
-# Copyright (C) 2022  Blue Brain Project, EPFL
+# Copyright (C) 2021-2024  Blue Brain Project, EPFL
 #
-# This program is free software: you can redistribute it and/or modify
-# it under the terms of the GNU General Public License as published by
-# the Free Software Foundation, either version 3 of the License, or
-# (at your option) any later version.
-#
-# This program is distributed in the hope that it will be useful,
-# but WITHOUT ANY WARRANTY; without even the implied warranty of
-# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-# GNU General Public License for more details.
-#
-# You should have received a copy of the GNU General Public License
-# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+# SPDX-License-Identifier: Apache-2.0
 
 """
 Synthesize a single neuron with global direction
--- a/examples/synthesize_single_neuron_y_direction.py
+++ b/examples/synthesize_single_neuron_y_direction.py
@@ -1,0 +1,99 @@
+# noqa
+
+# Copyright (C) 2022  Blue Brain Project, EPFL
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+"""
+Synthesize a single neuron
+==========================
+
+This example shows how to synthesize a single cell with simple parameters.
+"""
+
+import numpy as np
+from pathlib import Path
+import json
+
+import neurots
+
+
+def run(output_dir, data_dir):
+    """Run the example for generating a single cell."""
+    np.random.seed(42)
+    params = json.load(open(data_dir / "bio_params.json"))
+    # use trunk angle with y_direction awareness
+    params["basal_dendrite"]["orientation"] = {
+        "mode": "pia_constraint",
+        "values": {"form": "step", "params": [1.5, 0.25]},
+    }
+    params["apical_dendrite"]["orientation"] = {
+        "mode": "normal_pia_constraint",
+        "values": {"direction": {"mean": [0.0], "std": [0.0]}},
+    }
+
+    # Initialize a neuron
+    N = neurots.NeuronGrower(
+        input_distributions=data_dir / "bio_distr.json",
+        input_parameters=params,
+    )
+
+    # Grow the neuron
+    neuron = N.grow()
+
+    # Export the synthesized cell
+    neuron.write(output_dir / "generated_cell_orig.asc")
+
+    np.random.seed(42)
+
+    # Initialize a neuron
+    N = neurots.NeuronGrower(
+        input_distributions=data_dir / "bio_distr.json",
+        input_parameters=params,
+        context={"y_direction": [0.0, 1.0, 0.0]},
+    )
+
+    # Grow the neuron
+    neuron = N.grow()
+
+    # Export the synthesized cell
+    neuron.write(output_dir / "generated_cell_y.asc")
+
+    np.random.seed(42)
+
+    # Initialize a neuron
+    N = neurots.NeuronGrower(
+        input_distributions=data_dir / "bio_distr.json",
+        input_parameters=params,
+        context={"y_direction": [1.0, 0.0, 0.0]},
+    )
+
+    # Grow the neuron
+    neuron = N.grow()
+
+    # Export the synthesized cell
+    neuron.write(output_dir / "generated_cell_x.asc")
+
+    # the rotated neuron should be the same as original one
+    from morph_tool.transform import rotate
+
+    rotate(neuron, [[0, -1, 0], [1, 0, 0], [0, 0, 1]])
+    neuron.write(output_dir / "generated_cell_x_rot.asc")
+
+
+if __name__ == "__main__":
+    result_dir = Path("results_single_neuron")
+    result_dir.mkdir(parents=True, exist_ok=True)
+
+    run(result_dir, Path("data"))
--- a/examples/synthesize_single_neuron_y_direction.py
+++ b/examples/synthesize_single_neuron_y_direction.py
@@ -62,7 +62,7 @@ def run(output_dir, data_dir):
     N = neurots.NeuronGrower(
         input_distributions=data_dir / "bio_distr.json",
         input_parameters=params,
-        context={"y_direction": [0.0, 1.0, 0.0]},
+        #context={"y_direction": [0.0, 1.0, 0.0]},
     )
 
     # Grow the neuron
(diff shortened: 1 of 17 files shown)


arnaudon:
PS: this PR is based on https://github.com/BlueBrain/NeuroTS/pull/94, which we may want to merge first

neurots/generate/algorithms/basicgrower.py
```diff
@@ -51,19 +51,20 @@ def initialize(self):
...
-    def bifurcate(self, current_section):
+    def bifurcate(self, current_section, pia_direction=None):
         """When the section bifurcates two new sections are created.
 
         This method computes from the current state the data required for the
         generation of two new sections and returns the corresponding dictionaries.
 
         Args:
             current_section (neurots.generate.section.SectionGrowerPath): The current section.
+            pia_direction (numpy.ndarray): Direction of the pia if different from `[0, 1, 0]`.
```
adrien-berchet, line 62: It can also be a simple Python list, right?

neurots/morphmath/bifurcation.py
```diff
@@ -19,57 +19,75 @@
...
     theta1 = angles[3] / 2.0
 
-    dir1 = rt.rotate_vector(direction, [0, 0, 1], phi1)
-    dir1 = rt.rotate_vector(dir1, [1, 0, 0], theta1)
-    dir2 = rt.rotate_vector(direction, [0, 0, 1], -phi1)
-    dir2 = rt.rotate_vector(dir2, [1, 0, 0], -theta1)
+    x_dir = np.array([1, 0, 0])
+    z_dir = np.array([0, 0, 1])
+    if pia_direction is not None:
+        rot = rt.rotation_matrix_from_vectors(PIA_DIRECTION, pia_direction).T
```
adrien-berchet, line 44: As far as I understand, the `pia_direction` is only used in this file, so what about computing this rotation matrix earlier and then pass it through all the functions instead of the value of `pia_direction` so we don't compute the rotation matrix again for each bifurcation?

neurots/generate/algorithms/basicgrower.py
arnaudon: yes, I can add in the type

neurots/morphmath/bifurcation.py
arnaudon: yes, that could be better indeed, I'll try after lunch!

arnaudon: much cleaner this way!

arnaudon:
thanks @adrien-berchet , @lidakanari can we merge this?

eleftherioszisis:
Shouldn't the `pia_direction` be passed from the context or some other abstraction? You are passing the argument `pia_direction` to classes that are agnostic of a spatial embedding.

arnaudon:
it is passed via parameters file

neurots/generate/algorithms/tmdgrower.py
```diff
@@ -61,6 +65,12 @@ def __init__(
         self.apical_section = None
         self.apical_point_distance_from_soma = 0.0
         self.persistence_length = self.barcode.get_persistence_length()
+        if pia_direction is None:
+            self.pia_rotation = np.eye(3)
+        else:
+            self.pia_rotation = rotation.rotation_matrix_from_vectors(
+                PIA_DIRECTION, pia_direction
+            ).T
```
adrien-berchet, line 73: Would it make sens to move this in `AbstractAlgo.__init__` to avoid duplication?

neurots/generate/orientations.py
```diff
@@ -260,7 +260,7 @@ def _mode_normal_pia_constraint(self, values_dict, tree_type):
 
             phis = self._rng.uniform(0, 2 * np.pi, len(means))
             angles += spherical_angles_to_pia_orientations(
-                phis, thetas, self._parameters.get("pia_direction", None)
+                phis, thetas, self._parameters[tree_type].get("pia_direction", None)
```
adrien-berchet, line 263: Why is this change needed?

neurots/generate/algorithms/tmdgrower.py
arnaudon: I was thinking not to, so that growers can completely avoid using this (like astrocytes)

eleftherioszisis:
> it is passed via parameters file

You are passing `pia_direction` and `pia_rotation` to functions that should know nothing about a spatial context. Some of them are purely mathematical. Maybe you meant to have `global_orientation` or something like that?

neurots/generate/orientations.py
arnaudon: yes, this one is weird, I vaguely remember I changed something, but I can't find in on region-grower, let me revert

arnaudon:
is `pia` bothering you? its legacy, I'm to lazy to change its name

neurots/generate/algorithms/tmdgrower.py
```diff
@@ -61,6 +65,12 @@ def __init__(
         self.apical_section = None
         self.apical_point_distance_from_soma = 0.0
         self.persistence_length = self.barcode.get_persistence_length()
+        if pia_direction is None:
+            self.pia_rotation = np.eye(3)
```
adrien-berchet, line 69: As far as I can see, all the functions just do nothing when `pia_direction is None`, so why not just setting it to `None`?

arnaudon: Indeed!

neurots/generate/algorithms/basicgrower.py
```diff
@@ -22,6 +22,8 @@
 from neurots.generate.algorithms.abstractgrower import AbstractAlgo
 from neurots.generate.algorithms.common import bif_methods
 from neurots.generate.algorithms.common import section_data
+from neurots.morphmath import rotation
+from neurots.utils import PIA_DIRECTION
```
lidakanari, line 26: Maybe pia direction is too cortical, what about growth_direction?

neurots/generate/tree.py
```diff
@@ -140,6 +140,7 @@ def _initialize_algorithm(self):
             start_point=self.point,
             context=self.context,
             random_generator=self._rng,
+            pia_direction=self.params.get("pia_direction"),
```
lidakanari, line 143: I see this is exposed at the tree level only. This should be at the neuron level if the whole neuron is rotated?

lidakanari (commented):
You need to add a global test to generate a neuron at a couple of different directions and ensure the results do not change (i.e. the persistence and the local/global angles should remain the same). 

Also, I think it would be useful to expose this functionality at the neuron level. 

I am not sure if the results are correct as I cannot test on the population through the growth process.

neurots/generate/algorithms/basicgrower.py
arnaudon: principal_direction?

neurots/generate/tree.py
arnaudon: it is also use to place trunks, but from previous PRs

lidakanari:
My main concern is that the 'pia_direction' does not correspond to the orientation that is set as input for the apical. This could potentially cause an issue if one of the trees needs to grow towards a different direction, while the others grow  to other directions. 

Since this directionality is only exposed at the tree level, it is not obvious how the growth of the neuron as the whole will be affected. Shouldn't this impact be global rather than local at the tree level?

In addition, the global orientation has not been an issue for any other cell type apart from "flat" cells. Was this wrong so far but we missed it in the morphometrics? I am not sure how this correction will impact the pyramidal cells for example that are currently growing as expected taking into account the computed angle distributions.

neurots/generate/algorithms/basicgrower.py
lidakanari: sure! ah, this was introduced in the trunk modification?

neurots/generate/algorithms/basicgrower.py
```diff
@@ -36,10 +38,24 @@ class TrunkAlgo(AbstractAlgo):
...
     """
 
-    def __init__(self, input_data, params, start_point, context=None, **_):
+    def __init__(
+        self,
+        input_data,
+        params,
+        start_point,
+        context=None,
+        pia_direction=None,
```
lidakanari, line 47: Ok, correction to the previous comment. I see that this passes at the neuron level through the parameters. In this case, I would ask why you need to pass it outside here. If I understand correctly it is at a higher level so it is not accessible from the input parameters here (at the tree level). Is this correct? If that's the case, then it makes sense to add here.

arnaudon:
@lidakanari so I did a complete refactoring, it was indeed not making so much sense. I did the following:

- renamed PIA_DIRECTION to Y_DIRECTION
- renamed pia_direction to y_direction
- I pass pia_direction via the context, as it really is a context dependent parameter, it simplifies a lot of things in fact
- I added an example script (which I can move to the tests once this is agreed solution) that computes a neuron without this option, with this option set to default [0, 1, 0] value and with value [1, 0, 0].
- I made sure the resulting neuron when grown with y_direction = [1, 0, 0] is exactly the same as default one, when rotated from x to y.
- this works only with the trunk angle algo I implemented a while ago, the original one I didn't touch for this, as I don't use it via region-grower
- For this, I had to also rotate random numbers in the growth of section, which may be add a bit of computational cost for nothing, except ensuring neurons are the same, regardless of pia_direction value. We may not want this level of 'rotation invariance', but as you want.

I hope this is more clear now!

examples/synthesize_single_neuron_y_direction.py
```diff
@@ -0,0 +1,101 @@
...
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
```
adrien-berchet, line 16: This should be Apache 2 license, not GPL. You can copy a license header from `examples/synthesize_neurons.py`.

neurots/generate/algorithms/abstractgrower.py
```diff
@@ -23,7 +23,7 @@ class AbstractAlgo:
 
     def __init__(self, input_data, params, start_point, context):
         """The TreeGrower Algorithm initialization."""
-        self.context = context
+        self.context = context if context is not None else {}
```
adrien-berchet, line 26: In the existing functions we use statements like `if self.context is not None and ...` so you should make it consistent either by fixing all existing `if` statements or by fixing you new statements to handle the `None` case.

neurots/generate/section.py
```diff
@@ -97,6 +97,9 @@ def _propose(self, extra_randomness=0, add_random_component=True):
 
         if add_random_component:
             random_component = self.params.randomness * get_random_point(random_generator=self._rng)
+            # this is needed only to get 100% invariance under y_direction
+            # if self.context.get("y_rotation") is not None:
+            #    random_component = self.context["y_rotation"].dot(random_component)
```
adrien-berchet, line 102: Why is it commented?

examples/synthesize_single_neuron_y_direction.py
arnaudon: ah weird, I'll fix

neurots/generate/algorithms/abstractgrower.py
arnaudon: I'm not sure how to fix in fact, as the condition after None assume a dict with a specific key, so if we have {} instead of None, the if will work the same, no?

neurots/generate/section.py
adrien-berchet: I see. Maybe it would be nice to make the comment above a bit more intelligible then.

neurots/generate/algorithms/abstractgrower.py
adrien-berchet: I didn't say it can't work, I just said it's not consistent :)
If you ensure the context can never be `None`, then you should remove all the `if self.context not is None ...`. But this argument is still a bit experimental and there is not clear contract associated to it, so it's a bit hard to know how to handle it. For now I think it's always supposed to be a `dict` with some arbitrary items, but it would be better to create a specific class for this context object. But that's for our successors :smile:

arnaudon: ok, I removed the is None checks

neurots/generate/section.py
arnaudon: its difficult to make it invariant under noise, and not useful, so I removed this

neurots/generate/orientations.py
```diff
@@ -59,7 +59,7 @@ def __init__(self, soma, parameters, distributions, context, rng):
         self._soma = soma
         self._parameters = parameters
         self._distributions = distributions
-        self._context = context
+        self._context = context if context is not None else {}
```
lidakanari, line 62: Why is this different from the correction of context in grower?

self.context = self._process_context(context)

neurots/morphmath/bifurcation.py
```diff
@@ -10,55 +10,69 @@
 from neurots.morphmath.utils import get_random_point
 
 
-def random(random_generator=np.random):
+def random(random_generator=np.random, y_rotation=None):  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
arnaudon: because these functions need to have the same argument, they are called the same way in the code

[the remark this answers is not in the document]
lidakanari: ok!

neurots/generate/orientations.py
lidakanari: Indeed, if NeuronGrower has already processed it can it be None at this point?

arnaudon: yes, right, I put this to be sure it does not crash if used alone at all (like in the tests), as the context has to be a dict, as we use .get(). But default to {} in the parameter is dangerous.

arnaudon: @adrien-berchet maybe there is a better way to do this?

lidakanari: Yep, makes sense.  I agree with the default {} not being a good solution. I don't think this is critical, if you have a better alternative ok otherwise it's ok as it is.

adrien-berchet: Yeah I think it's ok like this

lidakanari (approved):

Issue #103, opened 2024-04-12
[Bug report]  The grow_to_target function used by AstrocyteGrower class has a an infinite loop


jacquemi-bbp:
### Describe the bug

The grow_to_target function astrocyte/section.py#L35  function used by AstrocyteGrower class has a an infinite loop https://github.com/BlueBrain/NeuroTS/blob/d47bf8d6448ad8290021ae8a24e69eaae6dbb203/neurots/

It is because one always adds the same reference to the point  object in a list instead of a new point object.
https://github.com/BlueBrain/NeuroTS/blob/d47bf8d6448ad8290021ae8a24e69eaae6dbb203/neurots/astrocyte/section.py#L68C8-L68C29

### To Reproduce

```Python
I can provide a script and the needed inputs to reproduce the bug,
```

### Error

```Python traceback
infinite loop
```

### neurots Version in Use

NeuroTS==3.5.0

### Python Version

3.10

### Operating system

Linux

jacquemi-bbp:
I will propose a PR to correct this BUG.

eleftherioszisis:
Please provide a script to reproduce the bug. :)

PR #104, opened 2024-04-12
Fix: Astrocyte grower infinite loop 

jacquemi-bbp:
- append a copy of point into the points list instead of a refence to the point.

### Description
Fix [103](https://github.com/BlueBrain/NeuroTS/issues/103)

### Checklist

This pull request is:

- [ ] A documentation / typographical error fix
	- Good to go, no issue or tests are needed
- [ ] A short code fix
	- [x] Please include: `Fixes: #<issue number>` in the description if it solves an existing issue
	  (which must include a complete example of the issue).
	- [ ] Please include tests that fail with the `main` branch and pass with the provided fix.
- [ ] A new feature implementation or update an existing feature
	- [ ] Please include: `Fixes: #<issue number>` in the description if it solves an existing issue
	  (which must include a complete example of the feature).
	- [ ] Please include tests that cover every lines of the new/updated feature.
	- [ ] Please update the documentation to describe the new/updated feature.


neurots/astrocyte/section.py
```diff
@@ -65,7 +65,7 @@ def grow_to_target(start_point, start_direction, target_point, segment_length, p
         )
 
         point += segment_length * direction
-        points.append(point)
+        points.append(point.copy())
```
eleftherioszisis, line 68: I would prefer if we didn't mutate that point in-place. So instead:
```python
point = point + segment_length * direction
points.append(point)
```
And avoid the in-place mutation altogether.

jacquemi-bbp:
How does the versioning of this package works ?
Do I need to change something for the versioning ?

eleftherioszisis:
> How does the versioning of this package works ? Do I need to change something for the versioning ?

No, I don't think so. Versioning should be automatic when a new release is made.

arnaudon (approved):

arnaudon:
write `Fix:blabla` in commit title

arnaudon:
ah, this changes the algo a bit apparently, as a test break, I guess you can update the test, it should not be such a big deal, but @eleftherioszisis should know more

eleftherioszisis (approved):

PR #112, opened 2024-11-30
Fix: Fix AR algo

arnaudon:


neurots/generate/orientations.py
```diff
@@ -234,24 +234,42 @@ def _mode_normal_pia_constraint(self, values_dict, tree_type):
...
-            means = means if isinstance(means, list) else [means]
-            stds = values_dict["direction"]["std"]
-            stds = stds if isinstance(stds, list) else [stds]
-            thetas = []
-            for mean, std in zip(means, stds):
-                if mean == 0:
-                    if std > 0:
-                        thetas.append(np.clip(self._rng.exponential(std), 0, np.pi))
+
+            def propose(_):
```
adrien-berchet, line 238: Seems the `propose()` and `prob()` functions could be defined outside of the loop, can't they?

arnaudon: better like that?

adrien-berchet (approved):
Is it still a draft?

